PR #2, opened 2026-03-12
Add operation serialization to VMLifecycleCoordinator

nicholas-lonsinger:
## Summary
Implements operation serialization in `VMLifecycleCoordinator` to prevent concurrent lifecycle operations on the same VM. Each VM can have at most one in-flight operation at a time, with concurrent requests rejected via a new `LifecycleError.operationInProgress` error.

## Key Changes

- **Token-Based Operation Serialization**: Added `activeOperations: [UUID: UUID]` dictionary mapping VM IDs to unique operation tokens
  - `serialized()` method wraps async operations with token acquisition/release
  - Token-based `defer` guard only cleans up if its token still matches, preventing stale operations from clobbering new entries
  - `hasActiveOperation(for:)` public method allows checking operation status

- **Stop/ForceStop Bypass Serialization**: `stop` and `forceStop` are intentionally not wrapped in `serialized()` — they are interruption/emergency operations that must always succeed
  - They call `activeOperations.removeValue(forKey:)` *before* invoking the service, invalidating the in-flight operation's token so its `defer` guard becomes a no-op

- **Serialized Lifecycle Methods**: `start`, `pause`, `resume`, `save`, `installMacOS`

- **New Error Type**: Added `LifecycleError` enum (nested inside `VMLifecycleCoordinator`) with `operationInProgress` case providing user-friendly error messages

- **Comprehensive Test Coverage**: Tests covering:
  - Initial state verification and active operation detection during in-flight operations
  - Rejection of concurrent operations on the same VM
  - Allowance of concurrent operations on different VMs
  - Lock release after success and failure
  - Stop and forceStop bypass serialization even during active operations
  - Token prevents stale `defer` from clobbering after stop clears entry (the critical race condition test)
  - macOS installation forwarding and error handling

- **Test Infrastructure**: Created `SuspendingMockVirtualizationService` with precondition and scheduling assumption documentation, enabling tests to suspend operations mid-flight via `CheckedContinuation` and inspect state before resuming

## Implementation Details

- Since `VMLifecycleCoordinator` is `@MainActor`, no locks are required — a simple `[UUID: UUID]` dictionary is sufficient for tracking active operations
- Each operation generates a unique `UUID` token stored in the dictionary; the `defer` block only removes the entry if the token still matches, which prevents a stale operation's cleanup from wiping a subsequent operation's tracking
- `stop`/`forceStop` clear the token before calling the service, so they always succeed and any suspended operation's `defer` guard becomes a no-op
- Detailed logging at debug and warning levels for operation lifecycle tracking
- Architecture documentation updated to reflect serialization behavior

https://claude.ai/code/session_01YXkAZuJitdxMPL2NQPSn9j

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -153,7 +153,7 @@ All service implementations conform to protocols defined in `Services/Protocols/
 
 - **`VMLibraryViewModel`** is the central `@Observable` view model. It owns the array of `VMInstance`s and handles list-level operations: add, remove, rename, selection tracking. For lifecycle operations (start, stop, install), it delegates to `VMLifecycleCoordinator`.
 
-- **`VMLifecycleCoordinator`** is an `@MainActor` coordinator that owns the lifecycle services (`VirtualizationService`, `MacOSInstallService`, `IPSWService`). It orchestrates multi-step operations like macOS installation (which involves IPSW download → platform file creation → VM configuration → installation). This separation keeps `VMLibraryViewModel` focused on list management. The coordinator enforces **per-VM operation serialization** — at most one lifecycle operation can be in flight for a given VM at any time; concurrent requests are rejected with `VMLifecycleCoordinator.Error.operationInProgress`. `stop` and `forceStop` are interruption-aware and bypass serialization so users can always cancel hung operations.
+- **`VMLifecycleCoordinator`** is an `@MainActor` coordinator that owns the lifecycle services (`VirtualizationService`, `MacOSInstallService`, `IPSWService`). It orchestrates multi-step operations like macOS installation (which involves IPSW download → platform file creation → VM configuration → installation). This separation keeps `VMLibraryViewModel` focused on list management. The coordinator enforces **per-VM operation serialization** — at most one lifecycle operation can be in flight for a given VM at any time; concurrent requests are rejected with `VMLifecycleCoordinator.LifecycleError.operationInProgress`. `stop` and `forceStop` bypass serialization entirely (clearing the active-operation token before calling the service) so users can always cancel hung operations.
 
 - **`VMCreationViewModel`** drives the multi-step creation wizard. It tracks the current step, validates inputs at each stage, and produces a `VMConfiguration` + disk image on completion.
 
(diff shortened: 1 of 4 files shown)


nicholas-lonsinger (commented):
## Code Review

### Critical: Race condition between `stop`/`forceStop` bypass and `serialized()` defer

The `serialized()` method uses `defer { activeOperations.remove(instance.id) }` to clean up. When `stop` bypasses serialization and clears the flag, the original operation's `defer` still fires later and silently removes a *new* operation's tracking entry.

**Scenario:**
1. `start(vm)` → `serialized` inserts `vm.id`, suspends in `body()`
2. `stop(vm)` → bypasses serialization, removes `vm.id`
3. `start(vm)` again → `serialized` sees no active op, inserts `vm.id`, suspends
4. Step 1's task resumes → `defer` fires → removes `vm.id` **which now belongs to step 3's operation**
5. Step 3's operation is now running untracked — a third concurrent operation would be allowed

**Fix:** Use a token-based approach instead of a bare `Set<UUID>`:

```swift
private var activeOperations: [UUID: UUID] = [:] // vmID → operationToken

private func serialized<T>(_ instance: VMInstance, action: String, body: () async throws -> T) async throws -> T {
    guard activeOperations[instance.id] == nil else { throw Error.operationInProgress(vmName: instance.name) }
    let token = UUID()
    activeOperations[instance.id] = token
    defer {
        if activeOperations[instance.id] == token {
            activeOperations.removeValue(forKey: instance.id)
        }
    }
    return try await body()
}
```

This way the defer only removes *its own* entry, not one inserted by a subsequent operation.

---

### Minor issues

1. **`Error` shadows `Swift.Error`**: The nested `enum Error` shadows the `Swift.Error` protocol. While it works because it conforms to `LocalizedError`, it can cause confusing diagnostics if you ever need to reference `Swift.Error` in the same scope. Consider renaming it to `LifecycleError` (which is what the PR description already calls it).

2. **Misleading test names**: `stopLockReleasedAfterCompletion` and `stopLockReleasedAfterError` — `stop` never acquires a lock (it bypasses serialization and never inserts into `activeOperations`), so asserting `!hasActiveOperation` is vacuously true. These tests verify `stop` doesn't *add* an entry, not that it *releases* one. The names suggest the latter.

3. **Single-slot continuation in `SuspendingMockVirtualizationService`**: The doc comment correctly warns about the single-slot limitation, but the `allowsConcurrentOperationsOnDifferentVMs` test works around it by toggling `shouldSuspendOnStart = false`. This is fragile — if someone adds a test that suspends two operations simultaneously, the first continuation gets silently leaked with no error.

4. **PR description inconsistency**: The description references `LifecycleError.operationInProgress` and mentions a `serializedSync()` method, but the implementation uses `VMLifecycleCoordinator.Error` and `stop` has no wrapper method — it calls the service directly and removes from the set inline. Please update the description to match the implementation.

PR #143, opened 2026-04-30
feat: Vsock guest <-> host channel + log forwarding

nicholas-lonsinger:
## Summary
- Introduce a vsock-based wire protocol between Kernova host and the macOS guest agent, sidestepping the SPICE/virtio-console reconciliation bug that produced `tty.com.redhat.spice.0X` orphans across save/restore on macOS guests
- Forward every macOS-guest agent log emission to the host so both sides of every clipboard exchange (and the agent's startup banner, SPICE handshake, disconnect events, etc.) are visible in one Console.app stream under `com.kernova.guest:<vmName>`
- Phases 1-3.5 of the migration plan; Phase 4 (clipboard transport over vsock, retire SPICE on macOS) is the next branch

## Changes
- New `KernovaProtocol/` Swift Package — first SPM dep on `apple/swift-protobuf`. Carries `Frame` / `Hello` / `Error` / `LogRecord` proto, a length-prefixed framing codec (`VsockFrame`, 4-byte BE prefix, 16 MiB cap), a bidirectional channel (`VsockChannel`), and 17 unit tests
- Host side — `ConfigurationBuilder` now appends a `VZVirtioSocketDeviceConfiguration` for macOS guests; `VsockListenerHost` wraps `VZVirtioSocketListener` on port 49153; `VsockGuestLogService` republishes incoming `LogRecord` frames under `com.kernova.guest:<vmName>` via `os.Logger`; lifecycle wiring in `VirtualizationService` / `VMInstance` / `ConfigurationBuilder`; 5 unit tests using `socketpair(AF_UNIX)` to simulate the wire
- Guest agent (macOS) — `VsockHostConnection` opens `socket(AF_VSOCK, SOCK_STREAM)` to `(VMADDR_CID_HOST, 49153)` with a 5s reconnect interval and a bounded 64-frame pre-connect log buffer flushed on Hello; `KernovaLogger` is a drop-in `os.Logger` replacement that preserves the `\(value, privacy: .public)` call-site syntax, emits to local `os.Logger`, and forwards every record over vsock
- The SPICE clipboard path is intentionally unchanged. Linux guests keep using it; the macOS-guest SPICE clipboard stays alive alongside the new vsock channel until Phase 4 retires it
- `CLAUDE.md` and `ARCHITECTURE.md` updated to reflect the refined "no third-party dependencies — Apple-published packages OK" policy and the structural additions (new directories, components, helper-target description)

## Notes
- Apple guards `OSLogMessage(stringInterpolation:)` against external construction and `OSLogPrivacy` can't be forwarded through as a runtime value (compile-time-literal-only). `KernovaLogger` therefore does its own redaction: `localRendered` substitutes `<private>` for `.private`/`.sensitive` values and is passed to `os.Logger` as a `.public` argument; `wireRendered` is cleartext and travels over vsock to the host (single-tenant trusted destination). Privacy contract holds — Console.app on the guest just shows our `<private>` literal rather than the OS's runtime-controlled placeholder.
- Every pre-existing `\(value, privacy: .public)` interpolation in the agent compiled unchanged after migration. Zero call sites needed editing.
- The vsock listener attaches before `vm.start()` so a fast-launching guest agent always finds the listener waiting.

## Test plan
- [x] `swift test` in `KernovaProtocol/` passes 17/17 (framing edge cases + channel round-trip / EOF / close via `socketpair`)
- [x] `xcodebuild -scheme Kernova … test` passes — host suite + 5 new `VsockGuestLogServiceTests`, no existing-test regressions
- [x] `xcodebuild -scheme Kernova … build` succeeds (compiles Kernova app, KernovaGuestAgent, KernovaRelaunchHelper, all SPM resolution clean)
- [x] Smoke-tested end-to-end on a real macOS 26.4.1 guest VM: agent installs from the bundled DMG, vsock connection establishes (~20-45s after VM start, depending on agent launch timing), the buffered startup banner flushes after Hello, SPICE handshake / clipboard grab+request+data / shutdown notice all forward to the host's `com.kernova.guest:Kernova Agent` log subsystem
- [ ] Linux-guest unaffected — Linux path is untouched (only macOS guests get the vsock device); not retested in this round but the host code branches on `config.bootMode == .macOS`


nicholas-lonsinger:
Addressed Gemini's review. Summary:

| Finding | Priority | Action | Commit / Issue |
|---------|----------|--------|----------------|
| `VsockChannel` lacks `deinit` — dropped instance hangs `incoming` consumers | high | **Fixed inline** — added `deinit { close() }`, leaning on the existing idempotent `close()` flag | `d2a3e7a` |
| `sockaddr_vm.svm_len` not set explicitly | medium | **Fixed inline** — set to `UInt8(MemoryLayout<sockaddr_vm>.size)` before `connect()` | `72ffcc3` |
| `openVsockToHost` uses blocking `connect`; `SO_*TIMEO` doesn't bound it on Darwin | medium | **Filed as follow-up** — non-blocking + select/poll is well-bounded work but mid-sized; vsock connects in practice are near-instant or `ECONNREFUSED`, so this is latent rather than observable | #146 (`review-debt/refactor`) |
| `VsockFrameDecoder.removeSubrange` is O(N) per frame extraction | medium | **Filed as follow-up** — fine for current log volumes; switch to read-offset + periodic compaction once Phase 4 routes clipboard payloads | #147 (`review-debt/performance`) |

Tests: 18/18 in the package, full host suite still TEST SUCCEEDED locally.

PR #148, opened 2026-04-30
feat: Clipboard sync over vsock (host + guest)

nicholas-lonsinger:
## Summary

- Replace the SPICE-based clipboard transport for macOS guests with a vsock-native one running on `KernovaVsockPort.clipboard` (49152). End-to-end: the host's `VsockClipboardService` and the guest's `VsockGuestClipboardAgent` exchange `ClipboardOffer` / `ClipboardRequest` / `ClipboardData` frames with monotonic generations and stale-frame drops.
- Linux guests continue to use the existing SPICE console-port path. The split is implicit in `VMInstance.startClipboardService()` — Linux runs `SpiceClipboardService`, macOS arms a vsock listener that lazily constructs `VsockClipboardService` when the guest connects.
- Reduces guest-agent surface: `GuestClipboardAgent.swift` and `SerialPortDiscovery.swift` (both SPICE-only) are gone, and the connect/retry/serve loop is extracted into a shared `VsockGuestClient` so log + clipboard channels share one implementation.

## Changes — host (commit `382ed01`)

- `kernova.proto`: add `ClipboardOffer/Request/Data/Release` (fields 20-23) + `ClipboardFormat` enum; regenerate Swift bindings.
- `KernovaVsockPort.clipboard = 49152`.
- New `ClipboardServicing` protocol; `SpiceClipboardService` conforms (no internal change). `VMInstance.clipboardService` becomes `(any ClipboardServicing)?` so consumers don't branch on transport.
- New `VsockClipboardService` — `@MainActor @Observable` class running the offer/request/data state machine over `VsockChannel`.
- `VMInstance.startVsockServices()` installs the clipboard listener alongside the log listener for macOS guests with `clipboardSharingEnabled`.
- `ConfigurationBuilder.configureClipboardSharing` now Linux-only (macOS clipboard rides the existing virtio-socket device).
- `VsockGuestLogService` switch updated to detect and log misrouted clipboard payloads on the log port.

## Changes — guest agent (commit `06e7b9c`)

- New `VsockGuestClient` — connect / retry / serve loop with `SO_RCVTIMEO`/`SO_SNDTIMEO` and the `sockaddr_vm.svm_len` fix from #143. Used by both log and clipboard.
- `VsockHostConnection` refactored to use `VsockGuestClient`; public surface (`start`, `stop`, `forwardLog`) unchanged.
- New `VsockGuestClipboardAgent` — polls `NSPasteboard.general` at 0.5 s, announces local changes via `ClipboardOffer`, requests on inbound offers, writes inbound data to `NSPasteboard`. State serialized on the main dispatch queue.
- `main.swift` starts both connections; the SPICE attempt loop is removed.
- `VsockPorts.swift` mirror on the agent side (deliberate duplicate per the existing comment in `VsockHostConnection` — port numbers are part of the wire contract).
- Delete `GuestClipboardAgent.swift` and `SerialPortDiscovery.swift`.
- Remove `SpiceAgentProtocol.swift`'s `KernovaGuestAgent` target-membership exception — the file is host-only now.

## Tests

- New `KernovaTests/VsockClipboardServiceTests.swift` — 8 cases via `socketpair`: hello, isConnected flip, monotonic offer generation, dedup/empty guards, request response, stale-request drop, inbound offer-driven population, stale-data drop.
- Full `xcodebuild test` passes.

## Test plan

- [x] `xcodebuild test` clean
- [ ] Smoke test in a real macOS guest VM:
  - [ ] Reinstall agent DMG, reboot guest
  - [ ] Confirm guest logs show "Vsock clipboard connected to host" and host logs show "Guest clipboard agent connected"
  - [ ] Copy text in the guest → host clipboard window populates
  - [ ] Type into the host clipboard window, lose focus → guest can paste it
  - [ ] Disconnect/reconnect (pause + resume VM) cycles cleanly

## Notes

- This branch deliberately bundles host + guest changes so the merged PR delivers a working clipboard. They are committed separately (`382ed01` and `06e7b9c`) for reviewability.
- The guest agent still emits `.debug` traffic on the log channel during clipboard activity; that's why the pre-connect log buffer is sized at 256.


nicholas-lonsinger:
Addressed all four Gemini findings, one fix per commit:

| Finding | Severity | Fix |
|---|---|---|
| Guest mutates dedup/generation state before send | HIGH | [1cac173](https://github.com/nicholas-lonsinger/kernova/commit/1cac173) — state updates moved inside the `do` block on both sides; on send failure state is left at pre-send so the next call retries |
| Dedup state not reset on reconnect | MEDIUM | [3b3d25c](https://github.com/nicholas-lonsinger/kernova/commit/3b3d25c) — `lastSeenText` cleared and `lastPasteboardChangeCount` reset to a sentinel inside the same dispatch that publishes the new `liveChannel`, so a fresh host instance receives the current pasteboard contents on reconnect |
| Host counter advances before send | MEDIUM | [1cac173](https://github.com/nicholas-lonsinger/kernova/commit/1cac173) — same as the guest fix above |
| Suppression test could miss spurious frames | MEDIUM | [d93f715](https://github.com/nicholas-lonsinger/kernova/commit/d93f715) — replaced ad-hoc `nextFrame` calls with a `FrameRecorder` pattern that owns a single iterator on the channel; suppression now asserted by checking the buffer is unchanged after each suppressed call |

Note on the test rewrite: the first attempt added an `expectNoFrame` helper that cancelled an iterator-reading task on timeout. `AsyncThrowingStream` is single-consumer, so cancelling its iterator terminates the shared iteration — every subsequent `nextFrame` call on the same channel then returned `nil`, masquerading as "channel finished". Switching to a recorder that owns one iterator for the test's lifetime sidesteps that.

Full suite green (`xcodebuild test`). Ready for another Gemini pass or merge.

PR #158, opened 2026-05-01
fix: Use non-blocking connect + poll for vsock reconnect (#146)

### LINKED ISSUE #146: Use non-blocking connect + select/poll for vsock reconnect
## Found during
PR #143 — Gemini code review (medium priority)

## Description
`VsockHostConnection.openVsockToHost(port:)` uses BSD blocking calls (`socket`, `connect`, `recv`/`send`). PR #143 added `SO_RCVTIMEO`/`SO_SNDTIMEO` at 30 s, which bounds **read/write** blocking — but on Darwin those options do *not* bound `connect(2)`. A wedged or never-listening peer therefore lets `connect` block until the kernel's own timeout (potentially much longer than the 5 s reconnect interval).

In practice this is rarely observable for vsock — local-only transport with no SYN dance, so connect is either immediate-success or immediate-`ECONNREFUSED`. But the unbounded blocking is a latent issue that becomes more visible under simulated kernel-level pathologies or once we use vsock for higher-volume traffic in Phase 4.

## Location
- `KernovaGuestAgent/VsockHostConnection.swift::openVsockToHost(port:)` — the `connect()` call inside `withUnsafePointer`
- The reconnect loop in `runReconnectLoop()` paces retries at 5 s but can't speed back up if a single `connect` blocks longer than that

## Suggested fix
1. After `socket()`, set `O_NONBLOCK` via `fcntl(fd, F_SETFL, O_NONBLOCK)`
2. Call `connect()` — expect `EINPROGRESS`
3. `select(2)` / `poll(2)` on the fd with the connect-side timeout (~3 s for vsock seems generous given local transport)
4. On timeout: close fd and let the reconnect loop schedule the next attempt
5. On select success: read `SO_ERROR` to determine connect outcome
6. Optionally clear `O_NONBLOCK` after success so subsequent `recv`/`send` follow the existing blocking-with-timeout model

This is mid-sized work (~30 lines) but well-bounded and not behavior-changing for the happy path.

## Notes
Not blocking for PR #143 — vsock connect failures observed in smoke testing have all been near-immediate.

nicholas-lonsinger:
## Summary
- Bounds `connect(2)` in `VsockGuestClient.openVsockToHost` so a wedged or never-listening host peer can't stall the reconnect loop longer than `connectTimeoutSeconds` (3 s)
- `SO_RCVTIMEO`/`SO_SNDTIMEO` only bound `recv`/`send` on Darwin — not `connect` — leaving the reconnect loop potentially blocked for the full kernel TCP timeout on a wedged peer
- Latent in practice (vsock connects are local and either immediate-success or immediate-`ECONNREFUSED`) but a real risk under kernel pathologies and Phase 4 high-volume traffic

## Changes
- Added `connectTimeoutSeconds = 3` constant alongside `socketTimeoutSeconds = 30`
- Rewrote `openVsockToHost` to use the non-blocking-connect-with-poll idiom: set `O_NONBLOCK` → `connect` → poll/`SO_ERROR` check on `EINPROGRESS` → restore blocking mode → apply `SO_RCVTIMEO`/`SO_SNDTIMEO`
- Extracted `awaitConnectCompletion(fd:label:port:)` helper — polls `POLLOUT` with 3 s timeout, retries on `EINTR`, reads `SO_ERROR` via `getsockopt` to surface deferred errors

## Test plan
- [x] Build succeeded on macOS 26
- [x] All 38 `KernovaGuestAgentTests` pass (including `VsockGuestClientTests` — 7 tests)
- [x] Full test suite passes (no regressions)
- [ ] Manual smoke: launch Linux/macOS guest, confirm log + clipboard channels connect and survive a host restart

Closes #146

### DIFF
--- a/KernovaGuestAgent/VsockGuestClient.swift
+++ b/KernovaGuestAgent/VsockGuestClient.swift
@@ -29,6 +29,7 @@ final class VsockGuestClient: @unchecked Sendable {
 
     private static let logger = Logger(subsystem: "com.kernova.agent", category: "VsockGuestClient")
     private static let socketTimeoutSeconds: Int = 30
+    private static let connectTimeoutSeconds: Int = 3
 
     let port: UInt32
     let label: String
@@ -136,7 +137,15 @@ final class VsockGuestClient: @unchecked Sendable {
 
     // MARK: - Socket helpers (static — no instance state read)
 
-    /// Opens a raw `AF_VSOCK / SOCK_STREAM` socket and connects to the host.
+    /// Opens a raw `AF_VSOCK / SOCK_STREAM` socket and connects to the host
+    /// using the non-blocking-connect-with-poll idiom so `connect(2)` can't
+    /// block the reconnect loop longer than `connectTimeoutSeconds`.
+    ///
+    /// `SO_RCVTIMEO`/`SO_SNDTIMEO` from Darwin do not bound `connect(2)`,
+    /// only `recv`/`send`. Non-blocking mode is used exclusively for the
+    /// connect phase; blocking mode is restored afterwards so subsequent
+    /// `recv`/`send` calls continue to observe the socket-level timeouts.
+    ///
     /// Returns the connected fd on success, nil on failure. Used as the
     /// default `socketProvider` in the production convenience init.
     private static func openVsockToHost(port: UInt32, label: String) -> Int32? {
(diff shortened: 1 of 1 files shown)


nicholas-lonsinger:
## Review findings addressed

All critical and important findings have been fixed in the latest commit (48a38f3).

### Fixed

**Critical**
1. **Stale `errno` in logging** — Captured `let err = errno` immediately after the failing `poll()` call and after the `getsockopt` failure path; both log sites now use the captured value.
2. **`pfd.revents` not inspected** — Added a check for `POLLHUP | POLLERR | POLLNVAL` output flags before trusting `SO_ERROR`. When error flags are set, `SO_ERROR` is still consulted for a more informative errno; if it's 0 or unavailable, the raw `revents` value is logged.

**Important**
3. **EINTR resets 3s budget** — Replaced the fixed `timeoutMs` with a `ContinuousClock` deadline; each retry computes the remaining time so the 3s ceiling is always respected regardless of how many EINTR signals arrive. Also captures `errno` into a local before the EINTR check.
4. **`fcntl` log levels** — Promoted all three `fcntl` failure log calls from `.warning` to `.error` per the CLAUDE.md level table (operations that did not complete = `.error`).
5. **`connectTimeoutSeconds` rationale** — Added `// RATIONALE: vsock is a local-only transport with no SYN dance…` comment explaining why 3s rather than 30s.
6. **Doc-comment cleanup** — Removed the duplicate "Restore blocking mode…" inline comment (function-level doc covers it), removed the "Snapshot the original flags…" comment, tightened the `awaitConnectCompletion` doc to surface the fd-ownership invariant and drop the redundant what-it-does prose.

### Deferred

7. **BSD-syscall failure paths not tested** — Filed as #159 (`review-debt/test-gap`). The five new failure branches in `openVsockToHost`/`awaitConnectCompletion` are above the `socketProvider` injection seam and would require a `RawSocketOps`-style abstraction or a connectable-but-half-open peer to exercise. Not blocking this PR.

PR #160, opened 2026-05-01
perf: Switch VsockFrameDecoder to a read-offset model

### LINKED ISSUE #147: VsockFrameDecoder: switch from removeSubrange to read-offset model
## Found during
PR #143 — Gemini code review (medium priority)

## Description
`VsockFrameDecoder.nextFrame()` consumes parsed frames by calling `buffer.removeSubrange(buffer.startIndex..<end)` on a `Data`. `Data.removeSubrange` shifts the remaining bytes, giving O(N) behavior per frame extraction where N is the unread portion of the buffer.

For the current use case (mostly small log frames at modest rate, plus occasional protocol housekeeping), this is not a bottleneck — and that's been called out in PR #143's own review as "fine for log volumes." But once Phase 4 routes clipboard payloads (potentially MB-scale image content) and any future high-throughput scenario emerges, the per-frame compaction cost will start to matter.

## Location
- `KernovaProtocol/Sources/KernovaProtocol/VsockFrame.swift::VsockFrameDecoder.nextFrame()` — the `buffer.removeSubrange(buffer.startIndex..<end)` call after a frame is extracted

## Suggested fix
Switch the decoder to a read-offset model:

1. Keep an `offset: Int` alongside `buffer: Data`
2. `nextFrame()` reads from `buffer[offset...]` and increments `offset` by frame size on success
3. When `offset` exceeds a threshold (e.g. half the buffer size, or a fixed `compactionThreshold = 64 * 1024`), copy the unread tail to the front and reset `offset = 0`
4. Existing public API (`feed`, `nextFrame`, `isEmpty`, `bufferedByteCount`) stays unchanged; only their semantics relative to the underlying storage shift

Alternative — pull in something like `swift-collections`' `Deque` if/when we have other reasons to depend on it, but that's a heavier change for a single hot-path fix.

## Suggested tests
- Decoder correctness already covered by `VsockFrameTests` — those should keep passing unchanged
- Add a benchmark or microbench-style test feeding many small frames and asserting decode wall-time scales linearly with frame count rather than quadratically with total bytes processed

## Notes
Not blocking for PR #143 — current frame volumes (log records under 1 KiB) make the O(N) cost vanishingly small. Worth scheduling alongside Phase 4 (clipboard payloads).

nicholas-lonsinger:
## Summary
- Replaces the `removeSubrange`-per-frame pattern in `VsockFrameDecoder` with a `readOffset` cursor that advances through the buffer on each successful extraction, turning per-frame extraction from O(unread) into O(1)
- Compaction is deferred and amortized: the consumed prefix is dropped at most once per 64 KiB consumed (or immediately when the buffer is fully drained via `removeAll(keepingCapacity: true)`, which is a count-reset with no copy)
- All public API (`feed`, `nextFrame`, `isEmpty`, `bufferedByteCount`) and their semantics are unchanged; the decoder is ready for MB-scale clipboard payloads in Phase 4
- Two new tests cover the compaction path: one verifies correctness across many compaction cycles (256 KiB), one verifies `bufferedByteCount` accuracy when a partial frame is left mid-stream after compaction

## Test plan
- [x] `swift test --package-path KernovaProtocol` passes (20/20 tests including 2 new)
- [x] Full Xcode test suite passes (`** TEST SUCCEEDED **`)

Closes #147

### DIFF
--- a/KernovaProtocol/Sources/KernovaProtocol/VsockFrame.swift
+++ b/KernovaProtocol/Sources/KernovaProtocol/VsockFrame.swift
@@ -122,7 +122,8 @@ public struct VsockFrameDecoder: Sendable {
         }
     }
 
-    /// Drops the consumed prefix when worthwhile. Cheap if no work is needed.
+    /// Two-tier compaction: drained buffers reset for free; partial buffers
+    /// shift only after the threshold to keep the per-byte copy cost amortized.
     private mutating func compactIfNeeded() {
         guard readOffset > 0 else { return }
         if readOffset == buffer.count {
--- a/KernovaProtocol/Sources/KernovaProtocol/VsockFrame.swift
+++ b/KernovaProtocol/Sources/KernovaProtocol/VsockFrame.swift
@@ -59,7 +59,14 @@ public enum VsockFrameError: Error, Sendable, Equatable {
 /// queue at a time.
 public struct VsockFrameDecoder: Sendable {
 
+    /// Compact the consumed prefix once the read offset crosses this many
+    /// bytes. Sized to amortize the shift cost (each byte is copied at most
+    /// once before being dropped) without keeping more than ~64 KiB of stale
+    /// storage live per decoder.
+    private static let compactionThreshold: Int = 64 * 1024
+
     private var buffer: Data = Data()
+    private var readOffset: Int = 0
 
     public init() {}
 
@@ -77,7 +84,8 @@ public struct VsockFrameDecoder: Sendable {
     ///   should be discarded — the buffer is left in place but the stream is
     ///   considered corrupt.
     public mutating func nextFrame() throws -> Data? {
-        guard buffer.count >= VsockFrame.lengthPrefixSize else { return nil }
+        let unread = buffer.count - readOffset
+        guard unread >= VsockFrame.lengthPrefixSize else { return nil }
 
         let payloadSize = Int(readLengthPrefix())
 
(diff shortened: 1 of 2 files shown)


nicholas-lonsinger:
Addressed all review findings:

**Fixed (findings 1–4):**
- Added `decoderPayloadIntegrityAcrossCompaction`: feeds enough small frames to cross the 64 KiB threshold, then splits the next frame mid-payload across two feeds, and asserts the recovered payload bytes match exactly. Catches off-by-one bugs in `unreadStart` math.
- Added `decoderDrainsAcrossCompactionBoundary`: bulk-feeds ~1,100 frames in a single `feed` call (well past the threshold) and drains all of them via successive `nextFrame()`, asserting each payload is correct. Exercises mid-drain compaction.
- Added `decoderPartialLengthPrefixAcrossCompaction`: crosses the threshold, then delivers only 1 byte of the next frame's 4-byte length prefix, verifies `nextFrame()` returns `nil`, then delivers the rest and checks payload integrity.
- Updated existing threshold-coupled byte-count comments (`// ~256 KiB of data`, `// ~128 KiB consumed`) to invariant-style: `// Enough frames to cross the compaction threshold several times.` / `// Cross the compaction threshold at least twice before leaving a partial frame.`

**Fixed (findings 5–6):**
- Replaced the `compactIfNeeded` doc-comment with: *"Two-tier compaction: drained buffers reset for free; partial buffers shift only after the threshold to keep the per-byte copy cost amortized."*
- Dropped the redundant `// Feed a half-frame to leave unread bytes hanging.` comment.

All 23 KernovaProtocolTests pass; no regressions in the full Kernova suite.

PR #161, opened 2026-05-01
fix: Use typed-error Result return in VsockSocketProvider to halt on permanent failures (#150)

### LINKED ISSUE #150: VsockSocketProvider: use typed-error return to short-circuit on permanent failures
## Found during
PR #149 review of `KernovaGuestAgent/VsockGuestClient.swift`

## Description
`VsockSocketProvider` currently returns `Int32?` — nil signals transient failure (retry), but there is no way for the provider to signal a permanent failure (e.g. `EAFNOSUPPORT` — AF_VSOCK not supported in this environment). The reconnect loop always sleeps and retries regardless, which wastes the retry interval and log output on every boot when vsock is unavailable.

A typed-error return (e.g. `Result<Int32, VsockProviderError>` where `VsockProviderError` has a `.permanent` case) would allow the loop to short-circuit immediately on permanent failures rather than retrying every 5 s indefinitely.

## Location
`KernovaGuestAgent/VsockGuestClient.swift` — `VsockSocketProvider` typealias and `connectAndServe` loop logic

## Suggested fix
Change `VsockSocketProvider` to `@Sendable (_ port: UInt32, _ label: String) -> Result<Int32, VsockProviderError>` and add a `.permanent` vs `.transient` discriminator so `runReconnectLoop` can break on permanent errors.

nicholas-lonsinger:
## Summary
- Replaces the `Int32?` return of `VsockSocketProvider` with `Result<Int32, VsockProviderError>` where `VsockProviderError` discriminates `.transient(String)` from `.permanent(String)`
- Permanent failures (`EAFNOSUPPORT`, `EPROTONOSUPPORT` from `socket()`) now exit the reconnect loop immediately with a single `.error`-level log, eliminating the one-warning-per-5s spam on kernels without AF_VSOCK support
- Transient failures (resource exhaustion, connect refused, poll/fcntl errors) preserve the existing sleep-and-retry behavior

## Changes
- `VsockGuestClient.swift`: add `VsockProviderError` enum and `LoopOutcome` private enum; update `VsockSocketProvider` typealias; rewrite `openVsockToHost` to return `Result`; add `classifySocketErrno` helper; reshape `connectAndServe` to return `LoopOutcome`; honor `.terminate` in `runReconnectLoop`
- `VsockGuestClientTests.swift`: update 7 provider closures to return `Result`; rename nil-retry test to reflect transient failure; add `permanentFailureHaltsLoop` test
- `VsockGuestClipboardAgentTests.swift`: update 3 provider closures (including `FdBox`-backed ones) to return `Result`

## Test plan
- [x] All existing tests pass (`** TEST SUCCEEDED **`)
- [x] `permanentFailureHaltsLoop` new test confirms provider called exactly once when permanent failure is returned, verifying the loop halts

Closes #150

### DIFF
--- a/KernovaGuestAgent/VsockGuestClient.swift
+++ b/KernovaGuestAgent/VsockGuestClient.swift
@@ -128,7 +128,13 @@ final class VsockGuestClient: @unchecked Sendable {
     private func runReconnectLoop(serve: @Sendable @escaping (VsockChannel) async -> Void) async {
         while !Task.isCancelled {
             let outcome = await connectAndServe(serve: serve)
-            if outcome == .terminate { break }
+            switch outcome {
+            case .terminate:
+                lock.withLock { stopped = true }
+                return
+            case .retry:
+                break
+            }
             guard !Task.isCancelled else { break }
             try? await Task.sleep(for: retryInterval)
         }
@@ -141,13 +147,14 @@ final class VsockGuestClient: @unchecked Sendable {
         switch socketProvider(port, label) {
         case .success(let f):
             fd = f
-        case .failure(.transient(let reason)):
-            Self.logger.warning("\(reason, privacy: .public)")
+        case .failure(.transient):
+            // Logged at the call site closest to errno context (openVsockToHost /
+            // classifySocketErrno). No re-log here — the typed error is purely a
+            // control-flow signal at this level.
             return .retry
-        case .failure(.permanent(let reason)):
-            Self.logger.error(
-                "\(reason, privacy: .public). Halting reconnect loop for '\(self.label, privacy: .public)'."
-            )
+        case .failure(.permanent):
+            // Logged at error level inside classifySocketErrno. Halt the loop.
+            Self.logger.error("Halting reconnect loop for '\(self.label, privacy: .public)' after permanent failure.")
             return .terminate
         }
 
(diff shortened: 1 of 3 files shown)


nicholas-lonsinger:
## Review findings addressed

All critical, important, and suggested findings from the review have been applied in 7d683b7.

**Critical (finding 1):** Restored `.error` log level for the three fcntl failure sites (lines 213, 219, 254 in VsockGuestClient.swift) — regression from #158 is corrected.

**Important (finding 2):** Eliminated duplicate logging. `connectAndServe`'s switch arms no longer re-log the reason string — the typed `VsockProviderError` is now a pure control-flow signal. Logs happen once, at the errno call site inside `openVsockToHost` / `classifySocketErrno`. A single `logger.error("Halting reconnect loop for '\(label)' after permanent failure.")` is retained in the `.permanent` arm as a lifecycle marker.

**Important (finding 3):** Added `ClassifySocketErrnoTests` suite with 5 tests: `EAFNOSUPPORT` → `.permanent`, `EPROTONOSUPPORT` → `.permanent`, `EMFILE` → `.transient`, `EACCES` → `.transient`, errno `0` → `.transient`. `classifySocketErrno` promoted from `private static` to `internal static`; the test target compiles agent sources directly so no `@testable import` is needed.

**Important (finding 4):** Added `startAfterPermanentTerminationIsNoOp` test — calls `start` a second time after the permanent-failure loop has exited, waits another retry window, and asserts `provideCounter.value == 1` still. Pins the "cannot be restarted" docstring contract.

**Suggestion (finding 5):** Added `lock.withLock { stopped = true }` in the `.terminate` case of `runReconnectLoop`, making terminal state explicit rather than relying on `reconnectTask` never being re-nilled.

**Suggestion (finding 6):** Replaced `if outcome == .terminate { break }` with an exhaustive `switch outcome` — compiler will now force handling of any future third case.

**Dismissed (per review):** `VsockProviderError` String payload refactor, typed throws migration, `EACCES`/`EPERM` → permanent reclassification, and `Equatable` conformance audit — all deferred as out of scope.

PR #162, opened 2026-05-01
fix: Guard setString return value in handleData (#151)

### LINKED ISSUE #151: Fix setString return value ignored in handleData corrupts echo-suppression state
## Found during
PR #149 review of `KernovaGuestAgent/VsockGuestClipboardAgent.swift`

## Description
In `handleData(_:)`, after calling `pasteboard.clearContents()` then `pasteboard.setString(text, forType: .string)`, the return value of `setString` is discarded. If `setString` returns `false` (write failed), the agent still updates `lastPasteboardChangeCount` and `lastSeenText` as though the write succeeded. This corrupts the echo-suppression state:

- `lastSeenText` is set to the host's text even though it was never written to the pasteboard
- On the next poll cycle, the agent sees a different `changeCount` (old pasteboard content) but `lastSeenText` points to the host text — echo suppression fires incorrectly, and the host never sees the guest's actual clipboard

## Location
`KernovaGuestAgent/VsockGuestClipboardAgent.swift` lines 341-347 (`handleData` method)

## Suggested fix
Guard on the `setString` return value: only update `lastPasteboardChangeCount` and `lastSeenText` when `setString` returns `true`. Log a warning and skip the state update on failure.

nicholas-lonsinger:
## Summary
- `handleData` now guards on `pasteboard.setString` return value; a `false` result logs a warning and returns early, leaving `lastSeenText` and `lastPasteboardChangeCount` at their pre-call values
- The success-path `.notice` log is moved past the guard so it only fires on a real write
- `FakePasteboard` gains `failNextSetString(times:)` for one-shot write-failure injection in tests
- Adds regression test `setStringFailureDoesNotCorruptEchoSuppression` that verifies the pasteboard stays empty and the next user-driven copy still produces an outbound offer

Fixes #151

## Test plan
- [x] All 7 `VsockGuestClipboardAgentTests` pass, including the new regression test

### DIFF
--- a/KernovaGuestAgent/VsockGuestClipboardAgent.swift
+++ b/KernovaGuestAgent/VsockGuestClipboardAgent.swift
@@ -351,7 +351,7 @@ final class VsockGuestClipboardAgent: @unchecked Sendable {
         pasteboard.clearContents()
         guard pasteboard.setString(text, forType: .string) else {
             Self.logger.warning(
-                "Failed to write host clipboard to pasteboard (\(text.count, privacy: .public) chars). Echo-suppression state preserved; next user clipboard change will offer normally."
+                "Failed to write host clipboard to pasteboard (gen=\(data.generation, privacy: .public), \(text.count, privacy: .public) chars). Echo-suppression state preserved; next user clipboard change will offer normally."
             )
             return
         }
--- a/KernovaGuestAgent/VsockGuestClipboardAgent.swift
+++ b/KernovaGuestAgent/VsockGuestClipboardAgent.swift
@@ -349,7 +349,12 @@ final class VsockGuestClipboardAgent: @unchecked Sendable {
         }
 
         pasteboard.clearContents()
-        pasteboard.setString(text, forType: .string)
+        guard pasteboard.setString(text, forType: .string) else {
+            Self.logger.warning(
+                "Failed to write host clipboard to pasteboard (\(text.count, privacy: .public) chars). Echo-suppression state preserved; next user clipboard change will offer normally."
+            )
+            return
+        }
         // Record so the polling timer doesn't echo this back to the host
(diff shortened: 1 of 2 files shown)


nicholas-lonsinger:
Addressed all review findings:

**Critical / Important (fixed now):**

1. **Regression test now catches the actual bug** — changed the user-copy text to be the same string the host failed to write. With the bug, `lastSeenText` was set to the failed text, so the subsequent `checkClipboardChange()` sees a match and echo-suppresses. With the fix, `lastSeenText` remains nil and the offer is sent. This is the first test that actually fails on the original (un-guarded) code.

2. **`FakePasteboard.clearContents()` now bumps `_changeCount`** — corrected to match real `NSPasteboard` behavior (which increments and returns the new count). Audited all tests; the bump shifts the absolute counts but no test relied on the count staying flat across `clearContents()`.

3. **Added `setStringFailureRetrySucceeds` test** — pins the recoverable-state property: after a failed write, `pendingInboundGeneration` is preserved; a second `ClipboardData` frame with the same generation succeeds, writes the pasteboard, and sets echo suppression correctly.

**Suggestions (applied):**

4. **Added `gen=` to the setString failure `.warning` log** — correlates inbound generation against host wire logs.

5. **`offer.generation >= 1` assertion** — left as-is; in `outboundOfferOnPasteboardChange` there is no prior offer value to compare against, so the `>= 1` is the meaningful assertion available. Dismissed as described in triage.

**PR comments (both addressed):**

- Initialized pasteboard with `"initial guest text"` before the failure test so `clearContents()` impact is observable and confirms `handleData` ran before asserting empty.
- Replaced `Task.sleep(for: .milliseconds(100))` with `try await waitUntil { pasteboard.string(...) == nil }` for deterministic synchronization on the failure path.

PR #164, opened 2026-05-01
fix: Send error frames on handleRequest failures in clipboard agents

### LINKED ISSUE #155: handleRequest send-data failure silently drops host clipboard request with no error response
## Found during
PR #153 review of `KernovaGuestAgent/VsockGuestClipboardAgent.swift`

## Description
When `handleRequest` fails to send `ClipboardData`, only a `.warning` is logged. The host has no retry mechanism for `ClipboardRequest` and no error frame is sent back — from the host's perspective it requested data and got nothing. Result: silent paste-doesn't-work. The user sees "I copied X on the guest, host shows the request landed, but the data never arrived" with only a `.warning` in the guest log and no signal to the host.

## Location
`KernovaGuestAgent/VsockGuestClipboardAgent.swift` — `handleRequest(_:channel:)`, the `catch` block around line 329.

## Suggested fix
On send-data failure, send a `Kernova_V1_Frame` with `.error` payload (`Kernova_V1_Error`) before returning, so the host can decide whether to retry or surface a user-visible "paste from guest failed" message. At minimum, include the generation number in the warning log and consider escalating to `.error` severity since this is user-observable.

nicholas-lonsinger:
## Summary
- Silent failure paths in `handleRequest` now reply with a `Kernova_V1_Error` frame so the peer knows the data transfer failed rather than waiting forever for data that will never arrive
- Escalates the data-send failure log from `.warning` to `.error` (this failure is user-visible: paste produces nothing)
- Both the guest agent (`VsockGuestClipboardAgent`) and the host service (`VsockClipboardService`) are fixed symmetrically — the host had the same four silent-failure shapes as the guest

## Failure paths fixed

| Path | Guest | Host |
|------|-------|------|
| Stale generation | Debug log only (silence is correct — peer has moved on) | Same |
| Unsupported format | `.debug` → `.warning` + error frame | Same |
| UTF-8 encoding failure | `.warning` + error frame | Same |
| `channel.send(dataFrame)` throws | `.warning` → `.error` (gen+bytes) + best-effort error frame | Same (was already `.error`, now also sends error frame) |

The `sendErrorFrame` helper is best-effort: if the channel died on the data send, the error frame attempt also fails and is swallowed with a `.debug` log. The peer learns via EOF in that case.

## Test plan
- [x] `KernovaGuestAgentTests/VsockGuestClipboardAgentTests` — 7 tests pass (new: `unsupportedFormatRepliesWithError`)
- [x] `KernovaTests/VsockClipboardServiceTests` — 10 tests pass (new: `unsupportedFormatRepliesWithError`)

Fixes #155

### DIFF
--- a/Kernova/Services/VsockClipboardService.swift
+++ b/Kernova/Services/VsockClipboardService.swift
@@ -52,6 +52,11 @@ final class VsockClipboardService: ClipboardServicing {
 
     private static let logger = Logger(subsystem: "com.kernova.app", category: "VsockClipboardService")
 
+    // Error codes for outbound `Kernova_V1_Error` frames on `handleRequest` failures.
+    private static let errorCodeFormatUnavailable = "clipboard.format.unavailable"
+    private static let errorCodeEncodingFailure = "clipboard.transfer.encoding.failure"
+    private static let errorCodeTransferFailure = "clipboard.transfer.send.failure"
+
     // MARK: - Init
 
     init(channel: VsockChannel, label: String) {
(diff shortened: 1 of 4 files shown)


nicholas-lonsinger:
## Review findings addressed

**Applied (all 7 in-scope items):**

1. **Path #3 tests added** — `dataSendFailureIsHandledGracefully` on both sides. Pattern: socketpair → `SO_NOSIGPIPE` on the writer's fd → queue the request in the kernel buffer → close the peer → agent/service processes the request, send fails, error frame also fails and is swallowed. Guest test asserts `liveChannelForTesting == nil` after peer EOF (state machine cleared). Host test asserts `isConnected` remains true (only `stop()` clears it, as designed).

2. **"Escalate severity" comments rewritten** — Both files now read: `// Logged at .error: this failure is user-visible (paste produces nothing). // Include gen + size so post-mortems can pair this with the peer's "request sent" log.`

3. **`gen=` assertion added** — Both `unsupportedFormatRepliesWithError` tests now assert `err.message.contains("gen=\(offer.generation)")`.

4. **`// Error codes for outbound...` block comments removed** — Both files.

5. **`sendErrorFrame` doc first sentence trimmed** — Both files; kept only the swallow-rationale paragraph.

6. **"Best-effort error frame" inline comment replaced** — Both files now: `// If the channel is dead, the peer will learn via EOF — no further fallback needed.`

7. **Narration comments removed** — Both test files. WHY comments (`// Expect an Error frame — not silence.` and `// any non-textUtf8 value`) preserved.

**Deferred (filed as review debt):**

- #165 — Promote `sendErrorFrame` to a shared `VsockChannel` extension in `KernovaProtocol` (review-debt/refactor)

All 47 tests pass.

PR #166, opened 2026-05-01
fix: Replace async liveChannel publish/cleanup with await MainActor.run in serve

### LINKED ISSUE #156: liveChannel async-publish vs post-loop cleanup race can expose a closed channel to the polling timer
## Found during
PR #153 review of `KernovaGuestAgent/VsockGuestClipboardAgent.swift`

## Description
In `serve`, `liveChannel = channel` is dispatched via `DispatchQueue.main.async` after Hello succeeds. If the host immediately closes the channel (network blip), the `for try await frame in channel.incoming` loop ends synchronously and the post-loop cleanup (lines ~163-169) dispatches its own `main.async` to clear `liveChannel`. If that cleanup async block runs before the publish async block, the publish then overwrites `liveChannel` with a now-closed channel. The polling timer then dispatches offers to a dead channel — a milder version of the same bug PR #153 fixes for Hello.

## Location
`KernovaGuestAgent/VsockGuestClipboardAgent.swift` — `serve(channel:)`, the two `DispatchQueue.main.async` blocks around lines 147-158 (publish) and 174-180 (cleanup).

## Suggested fix
Make the `liveChannel` publish happen synchronously before the read loop entry (e.g., with `await MainActor.run { ... }` if `serve` can `await`, or restructure to do the publish on main before starting the drain loop). Alternatively, gate the polling timer's sends on `liveChannel === <expected channel>` rather than just non-nil, which would at least prevent sends to a stale reference.

nicholas-lonsinger:
## Summary
- Closes a race where the polling timer could dispatch clipboard offers to a dead channel: when the host closed the channel immediately after Hello, the cleanup `DispatchQueue.main.async` could theoretically land on the main queue after the publish block. More concretely, the polling timer could tick between the async publish and cleanup, dispatching to a channel already closed by the peer.
- Both `DispatchQueue.main.async` blocks in `serve(channel:)` are replaced with `await MainActor.run { }`, making the publish synchronous before the read loop starts and the cleanup synchronous before `serve` returns. The reconnect loop in `VsockGuestClient.connectAndServe` can only advance past `await serve(channel:)` once serve returns, so `liveChannel` is guaranteed nil at that point.
- Frame-handler dispatches in the loop body and the `start`/`stop` blocks are unchanged — they don't need to block on anything.
- Host-side `VsockClipboardService` is `@MainActor @Observable` and owns its channel at init time with no dynamic channel swapping or polling timer, so the same race does not apply there.

Closes #156

## Test plan
- [x] New test `serveSynchronouslyClearsLiveChannelOnClose` passes — verifies `liveChannel` settles to nil after a host-close before any subsequent operation can observe stale state
- [x] All 51 KernovaGuestAgentTests pass
- [x] Full test suite passes (exit 0)
- [ ] Manual smoke: Linux guest clipboard sync over a host-app restart cycle

### DIFF
--- a/Kernova/Services/VsockClipboardService.swift
+++ b/Kernova/Services/VsockClipboardService.swift
@@ -28,6 +28,14 @@ final class VsockClipboardService: ClipboardServicing {
 
     // MARK: - Private state
 
+    // RATIONALE: `channel` is captured at init and never replaced. There is no
+    // polling timer or reconnect loop that reads the channel from a different
+    // actor context, so the publish/cleanup race fixed in
+    // VsockGuestClipboardAgent.serve (PR #166 / issue #156) does not apply here.
+    // If you add dynamic channel swapping (e.g., per-reconnect channel
+    // reassignment), mirror that fix: replace any DispatchQueue.main.async
+    // publish/cleanup in the new code with await MainActor.run so state settles
+    // before the function that changes the channel returns.
     private let channel: VsockChannel
     private let label: String
 
--- a/KernovaGuestAgent/VsockGuestClipboardAgent.swift
+++ b/KernovaGuestAgent/VsockGuestClipboardAgent.swift
@@ -159,6 +159,16 @@ final class VsockGuestClipboardAgent: @unchecked Sendable {
             return
         }
 
+        // RATIONALE: Use await MainActor.run rather than DispatchQueue.main.async
+        // here so the publish completes before serve() advances to the read loop.
+        // The polling timer checks liveChannel on the main queue; with async
+        // dispatch there would be a window where the timer fires, sees a nil
+        // channel, and silently skips an offer — even though serve() has already
+        // sent Hello and the connection is live. The other DispatchQueue.main.async
+        // sites in this class (start/stop, frame-handler dispatch in the loop body)
+        // are fire-and-forget and don't need this guarantee. MainActor.run shares
+        // the main-queue executor with DispatchQueue.main, so FIFO ordering is
+        // preserved between the two idioms.
         await MainActor.run {
             self.liveChannel = channel
             self.pendingOutbound = nil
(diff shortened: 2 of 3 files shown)


nicholas-lonsinger:
Addressed all review findings in the follow-up commit:

**Finding #1 (important) — Test pins synchronous invariant, not eventual consistency**
Replaced `serveSynchronouslyClearsLiveChannelOnClose` with a `DispatchSemaphore` gating pattern mirroring `helloFailureAbortsAndRetries`. The socket provider signals when it's called for the second fd (after the first connection closed and `serve()` returned); a background thread then snapshots `liveChannelForTesting` via `DispatchQueue.main.sync` *before* releasing the gate. The assertion checks that snapshot is `nil`. Under the old `DispatchQueue.main.async` cleanup, the snapshot could be non-nil because the async dispatch may not have run yet. Under the new `await MainActor.run`, cleanup completes before `serve` returns, so the snapshot is deterministically `nil`.

**Finding #2 (important) — RATIONALE on mixed concurrency idioms in serve()**
Added a `RATIONALE:` comment at the publish `await MainActor.run` block explaining why this site and the cleanup site use `await MainActor.run` while other sites in the class stay `DispatchQueue.main.async`, and noting that both idioms share the same main-queue executor so FIFO ordering is preserved. Added a second `RATIONALE:` at the cleanup block referencing the test that validates the invariant.

**Finding #3 (important) — RATIONALE on VsockClipboardService**
Added a `RATIONALE:` comment on the `channel` property in `VsockClipboardService` explaining that it's captured at init and never replaced, and that dynamic channel swapping would require mirroring the `await MainActor.run` fix.

**Finding #4 (suggestion) — Redundant `#expect` after `waitUntil`**
Gone naturally: the new gating test uses a snapshot assertion, not `waitUntil` + `#expect`.

**Finding #5 (suggestion) — Publish-side symmetry test**
Added `servePublishesLiveChannelBeforeReadLoop`: after the host Hello reply, `waitUntil { liveChannelForTesting != nil }` settles, then `DispatchQueue.main.sync` confirms non-nil, then an offer is sent and the agent's `ClipboardRequest` response confirms the read loop is running with `liveChannel` already set.

All tests pass (`** TEST SUCCEEDED **`).

PR #182, opened 2026-05-04
test: Harden vsock timing tests against macos-26 CI jitter

nicholas-lonsinger:
## Summary
- Bump `nextFrame` / `waitUntil` deadlines from 2 s to 5 s across the vsock test helpers — one prior post-merge failure (`VsockClipboardServiceTests/dropsUnsupportedProtocolVersion`) was a 2 s timeout firing on a slow macos-26 GitHub runner.
- Redesign `heartbeatOutboundCadence` to assert on inter-frame gap rather than count-in-window — that shape was the most jitter-sensitive in the suite and produced one prior `count → 2 ≥ 3` flake.
- Consolidate per-file `nextFrame` / `waitUntil` / `TestFailure` copies into a shared `TestHelpers.swift` per test bundle so both bundles emit the same diagnostic (`Timed out…` vs `EOF`) on any future flake.

## Changes
- Add `KernovaTests/TestHelpers.swift` with `@MainActor`-isolated `waitUntil` and `nextFrame`, plus shared `TestFailure` / `makeRawSocketPair` primitives. Mirrors the shape of the existing `KernovaGuestAgentTests/TestHelpers.swift`.
- Bump `waitUntil` / `nextFrame` / `awaitFirst` defaults in `KernovaGuestAgentTests/TestHelpers.swift` from 2 s to 5 s.
- Replace the count-in-window heartbeat assertion with a max-gap-over-3-consecutive-frames assertion (10× cadence tolerance) in both `VsockControlServiceTests.heartbeatOutboundCadence` and `VsockGuestControlAgentTests.heartbeatOutboundCadence`.
- Remove duplicate private `nextFrame` / `waitUntil` / `TestFailure` from `VsockClipboardServiceTests` and `VsockControlServiceTests`; tests now resolve to the bundle-shared `TestHelpers` definitions.
- Bump `SpiceClipboardServiceTests` pipe-EOF deadline from 1 s to 5 s.

## Test plan
- [x] `xcodebuild test` passes locally on macOS 26
- [ ] CI green on this PR
- [ ] `/gemini review` triaged

## Notes — deferred follow-ups
1. **Hardcoded post-action `Task.sleep` waits** in `VsockClipboardServiceTests` (lines 408 and 533): on closer reading, both test *negative state* (no crash; no spurious mutation). There's no positive event to poll for, so replacing with `waitUntil` would be a no-op or require invasive test seams.
2. **Injecting a `Clock` into `VsockControlService`** for fully deterministic liveness tests: Swift's `Clock` protocol has an associated `Instant` type, so doing this right needs either a generic class (invasive across call sites) or a custom mock clock that emits `ContinuousClock.Instant`. Saved for a focused refactor PR — the gap-based heartbeat assertion already removes the most-flaky shape.
3. **`VsockGuestControlAgentTests/inboundFramesAccepted` SIGPIPE crashes** seen in two pre-merge runs of #181: investigated and confirmed already fixed by the process-wide `signal(SIGPIPE, SIG_IGN)` static init added in `VsockChannel.swift:55-57` in the same PR. No code change needed.

### DIFF
--- a/KernovaGuestAgentTests/TestHelpers.swift
+++ b/KernovaGuestAgentTests/TestHelpers.swift
@@ -37,7 +37,7 @@ func makeChannelPair() throws -> (VsockChannel, VsockChannel) {
 
 /// Polls `predicate` every 10 ms until it returns `true` or `timeout` elapses.
 func waitUntil(
-    timeout: Duration = .seconds(2),
+    timeout: Duration = .seconds(5),
     _ predicate: @Sendable () -> Bool
 ) async throws {
     let deadline = ContinuousClock.now.advanced(by: timeout)
@@ -58,7 +58,7 @@ func waitUntil(
 ///   producing a frame (EOF), so the two failure shapes are identifiable.
 func nextFrame(
     from channel: VsockChannel,
-    timeout: Duration = .seconds(2)
+    timeout: Duration = .seconds(5)
 ) async throws -> Frame {
     let receiver = Task<Frame?, Error> {
         var iterator = channel.incoming.makeAsyncIterator()
@@ -87,7 +87,7 @@ func nextFrame(
 /// - Throws: `TestFailure("Timed out…")` if no value arrives within `timeout`.
 func awaitFirst<T: Sendable>(
     _ stream: AsyncStream<T>,
-    timeout: Duration = .seconds(2)
+    timeout: Duration = .seconds(5)
 ) async throws -> T {
     let task = Task<T?, Never> {
         var iterator = stream.makeAsyncIterator()
--- a/KernovaGuestAgentTests/VsockGuestControlAgentTests.swift
+++ b/KernovaGuestAgentTests/VsockGuestControlAgentTests.swift
@@ -86,27 +86,43 @@ struct VsockGuestControlAgentTests {
         host.start()
         defer { host.close() }
 
-        // Use a 100 ms cadence and a 1 s deadline so MainActor scheduling
-        // jitter on slow CI runners (the macOS-26 VMs see noticeably more
-        // latency than local hardware) doesn't squeeze the heartbeat count.
-        // 1 s / 100 ms = 10 expected cadences; demanding ≥ 3 leaves 7×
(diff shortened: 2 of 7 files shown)


nicholas-lonsinger:
Triaged gemini review:

- ✅ **medium / SpiceClipboardServiceTests:213 ad-hoc poll** — fixed in 8c2593d. Now uses the shared `waitUntil` helper.
- ❌ **high / TestHelpers.swift:81 receiver task leak** — false positive. `receiver.value` is awaited on every exit path (success, EOF, or timeout-induced `CancellationError`); the task always completes before the function returns, so a `defer { receiver.cancel() }` would be a no-op. Annotated with a `RATIONALE:` comment in 8c2593d so future reviews don't re-flag it. The same shape exists in `KernovaGuestAgentTests/TestHelpers.swift` and has been stable across many runs.
- ❌ **high / VsockGuestControlAgentTests:113 `.now` compile error** — false positive. `ContinuousClock.Instant.now` is a real Swift stdlib extension (`extension ContinuousClock.Instant { public static var now: Self }`). Verified by a clean local `xcodebuild test` pass on the parent commit.
- ❌ **high / VsockControlServiceTests:273 `.now` compile error** — same false positive as above.

The CI failure on the previous run was a transient `swift-protobuf` package fetch error (network), unrelated to the diff. This push should retrigger CI.

nicholas-lonsinger:
Both medium findings from gemini's incremental review fixed in b094310:

- **VsockGuestControlAgentTests:118** — explicit `timeout: .seconds(2)` removed; falls through to the 5 s default.
- **VsockControlServiceTests:269** — same.

Rationale: the explicit 2 s contradicted the PR's hardening goal. Worse, a real cadence drift > 2 s would throw a generic "Timed out" before the maxGap assertion could produce the sharper "Heartbeat cadence drift" diagnostic. Locally the redesigned tests still complete in ~0.31 s — the 5 s ceiling only matters in the failure case.

Both `heartbeatOutboundCadence` tests pass locally; pushing to retrigger CI.

PR #218, opened 2026-05-09
refactor: Promote KernovaProtocol to peer package so xcodebuild runs its tests

### LINKED ISSUE #215: Wire KernovaProtocolTests into make test and Periphery scan
## Found during
PR #214 — Gemini caught that I had deleted `VsockFrameDecoder.isEmpty` and `bufferedByteCount` despite their use in `VsockFrameTests.swift`. Local `make test` passed because it never runs the package tests.

## Description
The local `KernovaProtocol` SwiftPM package has its own test target (`KernovaProtocolTests`, 28 tests in 2 suites covering `VsockFrame` and `VsockChannel`), but neither **`make test`** nor the **Periphery dead-code scan** currently exercises it:

1. **`make test`** runs `xcodebuild test -scheme Kernova`. The `Kernova.xcscheme` `TestAction` lists `KernovaTests` and `KernovaGuestAgentTests`, but not `KernovaProtocolTests` (which lives inside the package, not the project). xcodebuild does not auto-discover SwiftPM package test targets — they have to be added to a scheme/test plan explicitly. Result: passing `make test` ≠ passing `swift test`. Confirmed in #214 where deleted public symbols still passed `make test` but failed `swift test` from `KernovaProtocol/`.

2. **Periphery scan** uses the same scheme list (`Kernova`, `KernovaGuestAgent`, `KernovaRelaunchHelper`). None of those schemes builds the package test target either, so Periphery indexes only the package's `Sources/`, never its `Tests/`. Result: any symbol used *only* by `KernovaProtocolTests` is reported as `warning: Unused property` even though it has live test coverage. This is exactly what tripped #214 (`VsockFrameDecoder.isEmpty` / `bufferedByteCount`).

## Location
- `Makefile:13` — `xcodebuild ... -scheme Kernova ... test`
- `Kernova.xcodeproj/xcshareddata/xcschemes/Kernova.xcscheme` — `<TestAction>` block
- `.github/workflows/xcodebuild-test.yml` — same `-scheme Kernova` invocation
- `.periphery.yml:11-14` — `schemes:` list

## Suggested fix
Two reasonable shapes — pick one (or both):

1. **Cheapest**: append a `swift test --package-path KernovaProtocol` step to `make test` and to `.github/workflows/xcodebuild-test.yml`. Keeps the SwiftPM tests as their own pass; doesn't touch Xcode test plans. Periphery would still miss these tests, but that can be addressed separately by adding the package's test target as a scheme testable, or by using `targets:` instead of `schemes:` in `.periphery.yml`.

2. **More integrated**: add the `KernovaProtocolTests` target as a `<TestableReference>` inside `Kernova.xcscheme`'s existing `<Testables>` block. `xcodebuild test -scheme Kernova` would then run it alongside the project tests, and Periphery (which uses the same scheme) would index it for symbol-graph analysis. Requires confirming Xcode lets you add a SwiftPM-package test target as a scheme testable.

Either fix retroactively justifies dropping the `// periphery:ignore` annotations added in #214 for `VsockFrameDecoder.isEmpty` / `bufferedByteCount`.

nicholas-lonsinger:
## Summary
- Switch `KernovaProtocol` from `XCLocalSwiftPackageReference` (Xcode "Add Package Dependencies → Add Local") to a peer `PBXFileReference` at the project root, so Xcode treats it as first-party rather than upstream.
- Side effect: `KernovaProtocolTests` now appears in `Edit Scheme → Test → +` and runs as part of `xcodebuild test -scheme Kernova` via the new `Kernova.xctestplan`.
- This eliminates the need for the dual-runner workaround introduced in #216 — `make test` and CI are back to a single `xcodebuild` invocation.

## Background — why this is the right shape
`xcodebuild test -scheme Kernova` was silently skipping `KernovaProtocolTests` (the package's 28-test SwiftPM `.testTarget`), which is what allowed PR #214 to delete `VsockFrameDecoder.isEmpty` and `bufferedByteCount` despite live test references.

#216 papered over the gap by adding a parallel `xcrun swift test --package-path KernovaProtocol` step to both `Makefile`'s `test` target and the CI workflow. That fixed gating but left the root cause in place: Xcode wasn't surfacing the package's test target in the scheme's test-plan picker.

The actual root cause: `KernovaProtocol` was added via *Add Package Dependencies → Add Local*, which produces an `XCLocalSwiftPackageReference`. Xcode treats those as upstream dependencies and intentionally hides their test targets from the scheme picker (the rationale being that you don't want to run third-party libraries' tests every time you `Cmd+U`). Re-adding the same folder via drag-from-Finder into the Project Navigator at the top level produces a regular `PBXFileReference` instead — Xcode then treats it as first-party, surfaces its test target in the picker, and lets it be added to the test plan.

## Changes
- `Kernova.xcodeproj/project.pbxproj`: drop `XCLocalSwiftPackageReference`; add `KernovaProtocol/` as a top-level `PBXFileReference` in the project's main group. The `XCSwiftPackageProductDependency` entries (the actual library link from each consumer target) are unchanged.
- `Kernova.xctestplan` (new): explicit test plan listing `KernovaTests`, `KernovaGuestAgentTests`, and `KernovaProtocolTests` (the last via `container:KernovaProtocol`).
- `Kernova.xcscheme`: switch `TestAction` from auto-create (`shouldAutocreateTestPlan = "YES"`) to reference `Kernova.xctestplan`.
- `Makefile`: drop the second `xcrun swift test --package-path KernovaProtocol` line from the `test` target and the rationale comment above it. Keep `make test-package` as a focused shortcut for iterating on package tests in isolation.
- `.github/workflows/xcodebuild-test.yml`: drop the `Test KernovaProtocol package` step now that `xcodebuild test-without-building` covers it.
- `.claude/CLAUDE.md`: rewrite the "Build & Test" paragraph to describe the single-runner reality and document why the peer-package layout matters (so future re-adds use drag-from-Finder, not "Add Local").

## Notes
- This solves the underlying issue from #215 / #216. Closes #215 properly — the previous fix solved the symptom (CI gating) via a parallel runner; this solves the cause (Xcode hiding the test target from the picker).
- Periphery still uses the same scheme list. With `KernovaProtocolTests` now built as part of the scheme, a follow-up dead-code scan should index it — at which point the `// periphery:ignore` annotations on `VsockFrameDecoder.isEmpty` / `bufferedByteCount` from #214 can be dropped. Deferred to a separate change after we confirm the next Periphery run picks them up.

## Test plan
- [x] `make clean && make test` runs all three test targets in a single `xcodebuild` invocation. xcresult shows `KernovaTests`, `KernovaGuestAgentTests`, and `KernovaProtocolTests` with 699 passed, 0 failed, 0 skipped.
- [x] `make test-package` still runs the package suite in isolation.
- [ ] CI on this PR exercises the simplified single-step flow.


nicholas-lonsinger:
Triage of Gemini's reviews:

- **HIGH (both reviews): "XCSwiftPackageProductDependency entries missing `package` property" / "must update to point to new PBXFileReference"** — **Dismiss.** False positive based on the assumption that all local packages use the `XCLocalSwiftPackageReference` shape. The `package` attribute on `XCSwiftPackageProductDependency` is only used when the package is referenced via `XCLocalSwiftPackageReference` or `XCRemoteSwiftPackageReference`. In the peer-package layout this PR adopts, Xcode resolves product names by scanning peer `PBXFileReference` entries that point at folders containing a `Package.swift`. Empirical proof: CI's `build-and-test` job builds the `KernovaProtocol` library and links it into all three consumer targets (`Kernova`, `KernovaGuestAgent`, `KernovaRelaunchHelper`), then runs all 705 tests (including the 28 in `KernovaProtocolTests`).
- **HIGH (first review only): "packageReferences list is empty"** — **Dismiss.** Same root misunderstanding. `packageReferences` is the list of `XCLocalSwiftPackageReference` / `XCRemoteSwiftPackageReference` entries by schema — it does not (and cannot) contain `PBXFileReference` entries. Removing it for a peer-package layout is correct.
- **MEDIUM (first review): "Enable parallelization for KernovaProtocolTests"** — **Fixed** in 299227b.

Merging on green CI.

PR #237, opened 2026-05-20
feat: Auto-start VM after wizard creation

nicholas-lonsinger:
## Summary
- Add a pre-checked "Start this VM after creation" toggle to the Review step of the new-VM wizard so users can jump straight from configuring a new VM into a running one.
- The default is on; unchecking leaves the VM in its post-creation state for the user to start manually later.

## Changes
- `VMCreationViewModel`: new `startAfterCreate: Bool = true` property.
- `ReviewStep`: switches to `@Bindable creationVM` and adds a `Toggle` bound to the new property at the bottom of the form.
- `VMLibraryViewModel.createVM`: after the new instance is appended, persisted, and selected, `await start(instance)` when the flag is set. Routing is unchanged: macOS-install VMs go through `installAndAutoBoot` (kicks off an install Task and returns immediately); Linux/EFI VMs await VZ start. Errors surface via the existing `presentError` path.
- `ARCHITECTURE.md`: surgical bump of the `VMCreationViewModel` test count and a one-phrase note about the new coverage.

## Test plan
- [x] `make lint` clean (also enforced by pre-push hook)
- [x] `make build` succeeds
- [x] `make test` — all suites pass, including:
  - `startAfterCreate defaults to true` (wizard view model)
  - `createVM auto-starts the new VM when startAfterCreate is true (default)`
  - `createVM does not auto-start when startAfterCreate is false`
  - `createVM does not auto-start when bundle creation fails`

## Notes
- The wizard's "Create" button still awaits `createVM` before dismissing. For macOS install (the typical wizard outcome) the auto-start path is essentially instant; for Linux it adds the brief VZ-start window before the sheet closes. This matches how the sidebar Start button already feels and keeps `createVM` testable without yielding to a fire-and-forget Task.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -139,7 +139,7 @@ KernovaTests/
 ├── VMToolbarManagerTests.swift          # Toolbar manager item creation and state update tests
 ├── VMConfigurationCloneTests.swift     # Clone-specific configuration tests
 ├── VMLibraryViewModelTests.swift       # 39 tests for the central view model
-├── VMCreationViewModelTests.swift      # 44 tests for the creation wizard
+├── VMCreationViewModelTests.swift      # 49 tests for the creation wizard
 ├── VMLifecycleCoordinatorTests.swift   # Coordinator orchestration tests
 ├── VMInstanceTests.swift               # Runtime instance behavior tests
 ├── ConfigurationBuilderTests.swift     # VZ configuration translation tests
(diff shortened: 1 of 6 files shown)


nicholas-lonsinger:
Triage of Gemini's review:

- **ARCHITECTURE.md `44 → 49` (both locations)** — kept as-is. The doc count was already stale: `git show main:KernovaTests/VMCreationViewModelTests.swift | grep -c '^    @Test'` returns **48** before this PR, so `48 + 1 = 49` is correct. Gemini inferred the delta from the diff alone, which doesn't show the pre-PR true count.
- **Centralize the new Toggle string** — dismissed. The project has no `Strings`/`L10n` enum and no `NSLocalizedString` usage; every other section/toggle/button title in `Kernova/Views/` is an inline literal. Adding a one-off centralization for this single string would be inconsistent with the rest of the UI.

Issue #342, opened 2026-06-13
Decide whether ConcealedType pasteboard content should be excluded from clipboard sync


nicholas-lonsinger:
## Summary

`ClipboardSnapshotPolicy` skips `org.nspasteboard.*` marker representations (TransientType, ConcealedType, AutoGeneratedType) — the markers are metadata, not content. But when a pasteboard item carries `org.nspasteboard.ConcealedType` (the convention password managers use to flag secrets), the *content itself* (the concealed text/data representations alongside the marker) still syncs to the guest.

This matches the pre-UTI behavior — the text-only pipeline never inspected marker types at all — so it was deliberately kept for behavior parity when the generic-representations work landed. Whether Kernova should suppress the entire snapshot when a ConcealedType marker is present is a product decision.

## Why

Privacy: a user copying a password from a password manager and then using the clipboard window (or the guest agent poll picking it up) may not expect the secret to cross into a VM. On the other hand, pasting a password *into* a guest is a common, legitimate workflow, and the clipboard window is already a manual, gated surface on the host side — the guest-agent poll is the path that picks content up without a per-item gesture.

## Considerations

- The guest-side poll (`VsockGuestClipboardAgent.checkClipboardChange()`) and the host window intake (`ClipboardPasteboardIntake`) share `ClipboardSnapshotPolicy`, so a decision lands in one place (e.g. a whole-snapshot suppressor in `shouldSkipBeforeReading` / `evaluate`).
- If suppressed, the skip should be user-visible (the policy's `Skipped` reporting and the window's transient-message strip already exist for this).
- TransientType / AutoGeneratedType whole-snapshot handling could be decided at the same time.

Issue #341, opened 2026-06-13
feat: Rich clipboard formats (images + UTI representations) over the SPICE/Linux transport


nicholas-lonsinger:
## Summary

The vsock/macOS clipboard pipeline now carries generic UTI-tagged representations (images, RTF, arbitrary data) — see the PR that closes #112. The SPICE transport for Linux guests is still text-only: `SpiceClipboardService.grabIfChanged()` is a no-op for content without a UTF-8 text representation, and `supportsBinaryRepresentations` is `false` so the clipboard window's intake rejects non-text content for Linux guests with an explanatory message.

This issue tracks lifting that limitation so Linux guests can exchange images (and map a sensible subset of the generic representations) over the SPICE agent protocol.

## Why

Feature parity with macOS guests. The SPICE wire format already defines the needed clipboard types (`SpiceClipboardType`: `png`, `bmp`, `tiff`, `jpg`) — the text-only behavior is host-side gating in `SpiceClipboardService`, not a protocol gap.

## Scope / considerations

- Map `ClipboardContent` representations onto the fixed SPICE type set (preference order png > tiff > jpg > bmp per the original #112), converting via `NSBitmapImageRep` where the pasteboard format and SPICE format differ.
- **Prerequisite — multi-chunk message reassembly:** `SpiceAgentParser` currently assumes each VDI chunk contains one complete VDAgent message, but `spice-vdagent` splits large messages into ~2 KB chunks (`VD_AGENT_MAX_DATA_SIZE`). Image payloads will always span chunks. Worth verifying whether >2 KB *text* from a Linux guest is already broken today for the same reason.
- The parser's 1 MB buffer cap and 1 MB max chunk size need revisiting for multi-megabyte payloads, alongside the perf concern recorded in #112's comments (double `subdata(in:)` copy per message and O(n) `removeSubrange` per chunk in `SpiceAgentParser.tryParseNext()`).
- `SpiceClipboardService` should flip `supportsBinaryRepresentations` to `true` and surface `lastTransferIssue` for failures, matching `VsockClipboardService`.

Implementation design deliberately omitted — design when picked up.

## Relationship to other issues

Supersedes the SPICE half of #112 (whose vsock half shipped). Independent of #82 (automatic passthrough) and #135 (lazy pull).

nicholas-lonsinger:
[...]

Issue #145, opened 2026-04-30
Harden vsock listener with per-VM auth before non-trusted guest workloads


nicholas-lonsinger:
## Found during
PR #143 (vsock guest <-> host channel + log forwarding) review

## Description
The vsock listener currently accepts any connection on `KernovaVsockPort.log` (49153) without authentication. For the current single-tenant macOS-on-macOS scenario this is acceptable — the guest is the user's own VM and the host trusts everything the guest sends. But once Phase 4 brings clipboard data over the same transport, and especially if the guest could ever run untrusted code, two attack surfaces open up:

1. **Log spam / log injection** — any process inside the guest with vsock access can connect to port 49153 and emit `LogRecord` frames that the host republishes under `com.kernova.guest:<vmName>` in its `os.Logger`. A malicious in-guest process could fill the host's log store with arbitrary content, possibly impersonating the real guest agent's category strings.

2. **Clipboard data exfiltration / injection (Phase 4)** — once the channel carries clipboard payloads, an in-guest non-agent process connecting to the listener could read every clipboard entry the host pushes, or push attacker-controlled content into the host pasteboard.

The `Hello` `capabilities` list is logged unredacted at `.notice` — a malicious peer advertising wild strings would also surface in the host log without redaction (cosmetic, not exploitable).

## Suggested fix
Implementation options were sketched in the original Phase 1 plan; the most practical first pass is **per-session token auth**:

1. Host generates a random token at VM-create time, persisted into the VM bundle
2. Token is delivered into the guest via a one-shot mechanism (cloud-init seed iso, a write-only virtio file, or installed once by `install.command` on the agent's first run)
3. Agent presents the token in the `Hello.agentInfo` (or a dedicated `auth_token` field on `Hello`)
4. `VsockGuestLogService` rejects connections whose `Hello` doesn't include a matching token, calling `channel.close()` and logging at `.warning`

Alternative: PSK-TLS over vsock is overkill but well-understood; defer unless the threat model graduates.

## Location
- `Kernova/Services/VsockListenerHost.swift` — `acceptDuplicatedFd(_:dupErrno:)` currently accepts any connection
- `Kernova/Services/VsockGuestLogService.swift` — `handle(frame:emitter:label:)` republishes Hello/LogRecord without verifying sender
- `KernovaProtocol/Proto/kernova.proto` — `Hello` would gain a token field

## Notes
[...]

PR #343, opened 2026-06-13
feat: Add image and generic clipboard data support for macOS guests

### LINKED ISSUE #112: feat: Add image clipboard support to SPICE clipboard pipeline
## Summary

Expand the SPICE clipboard pipeline to handle image types (`png`, `bmp`, `tiff`, `jpg`) in addition to the existing `utf8Text` support. This requires changes on both the host side (`SpiceClipboardService`, `ClipboardContentView`) and the guest agent.

This is sub-issue 4 of #71.

### 1. Host-side: `SpiceClipboardService`

Three methods are currently gated to `utf8Text` only:

- **`handleClipboardGrab(types:)`** — only requests data when `types.contains(.utf8Text)`. Expand to also request image types with a preference order (prefer `png` > `tiff` > `jpg` > `bmp`)
- **`handleClipboardRequest(type:)`** — only responds when `type == .utf8Text`. Expand to convert current clipboard content to the requested image format
- **`handleClipboardData(type:data:)`** — only processes `utf8Text`. Expand to handle image types

**New observable state:**
- Add a `clipboardImage: NSImage?` (or `Data` + type) property alongside `clipboardText`
- Update grab/request flow to handle mixed content (SPICE protocol sends available types in GRAB; receiver picks which to request)

### 2. Host-side: `ClipboardContentView`

- Show an `Image` view when `clipboardImage` is non-nil
- Allow drag-and-drop of images into the clipboard window (host-to-guest image transfer in manual mode)
- Show content type indicator (text vs. image) in the status bar
- Keep existing `TextEditor` for text content — switch between views based on current content type

### 3. Guest agent: image support

- Read image data from `NSPasteboard.general` using `NSPasteboard.PasteboardType.png`, `.tiff`, etc.
- Convert between SPICE clipboard types and `NSPasteboard` types
- Send `CLIPBOARD_GRAB` with image type list when guest clipboard contains an image
- Write received image data to `NSPasteboard.general` with appropriate UTType

### 4. Rebuild DMG

Update the embedded guest agent DMG with the new binary that supports image types.

### Modified files
- `Kernova/Services/SpiceClipboardService.swift` — expand type handling in `handleClipboardGrab`, `handleClipboardRequest`, `handleClipboardData`; add `clipboardImage` observable property
- `Kernova/Views/Clipboard/ClipboardContentView.swift` — image rendering, content type switching, drag-and-drop
- `KernovaGuestAgent/main.swift` — image type handling with `NSPasteboard` image types
- Guest agent DMG — rebuild with updated binary

## Dependencies

- **Depends on #115** (DMG packaging pipeline) and **#116** (guest agent core) — requires the guest agent target, working clipboard implementation, and DMG rebuild infrastructure
- **Independent of #82** (automatic clipboard passthrough) and **#117** (installation UX) — can be done in parallel. When both #82 and #112 are complete, automatic passthrough will naturally work for images too

## Notes

- `SpiceClipboardType` already defines all image types (`png = 2`, `bmp = 3`, `tiff = 4`, `jpg = 5`) — no wire protocol changes needed
- `SpiceMessageBuilder.buildClipboardGrab`, `buildClipboardRequest`, and `buildClipboardData` already accept any `SpiceClipboardType` — they are not gated to text. The gating is only in `SpiceClipboardService`'s message handlers
- macOS `NSPasteboard` natively supports PNG and TIFF. BMP and JPEG may require `NSBitmapImageRep` conversion

nicholas-lonsinger:
## Summary

Implements the headline feature of #112 — copy/paste of images, generic data, and **files** between host and guest — on the vsock/macOS pipeline, and rebuilds the clipboard window around it.

Closes #112, #344, #345. The Linux/SPICE half of #112 (which needs multi-chunk VDI reassembly and parser work before images can even fit through it) moves to #341; the ConcealedType product decision is #342.

## Wire format

The clipboard buffer is a `ClipboardContent`: ordered UTI-tagged representations — each optionally carrying a suggested **filename** for file payloads — mirroring one `NSPasteboardItem`, so macOS↔macOS transfers carry anything with full fidelity (plain text, RTF, PNG/TIFF, dynamic `dyn.*` types, arbitrary files). Proto changes are **additive only** (`ClipboardRepresentation` message with `uti`/`data`/`filename`; `utis` on offers/requests; `representations` on data; `ClipboardFormat` frozen as legacy):

| Sender → Receiver | Behavior |
|---|---|
| new → new | UTIs offered, echoed in the request, answered with `representations` |
| new → old | old agent sees `formats=[TEXT_UTF8]` (populated whenever a text rep exists) and pulls text via the legacy fields; image-only offers are silently ignored |
| old → new | new side detects the legacy shape and answers with the legacy `data` field |

Both directions are pinned by unit tests. The guest agent's `MARKETING_VERSION` bump (0.10.0 → 0.14.0) drives the existing outdated-agent update prompt.

`ClipboardSnapshotPolicy` (in the shared package, applied identically by the guest-agent poll and the host window intake) filters transient `org.nspasteboard.*` markers and file references, drops zero-byte reps, and caps sizes at 10 MiB/representation and 12 MiB total — headroom under the 16 MiB vsock frame limit, validated by a frame-budget test. The caps are re-enforced on the **receive** path too (`sanitizedForApply`), so a non-conformant peer can't land an over-cap blob on the pasteboard. Skips are never silent (logged + user-visible notes), and oversized content surfaces as `clipboard.transfer.too.large` / a transient message instead of a silent paste failure.

## Cross-boundary files

Any copied or dragged file crosses as the file itself — its bytes tagged with the content UTI and name, under the per-rep cap — so the receiver **materializes** a real local temp file (`ClipboardFileStaging`) and offers its `public.file-url`: a Finder paste creates the file, matching local macOS. An image additionally pastes **inline** (Notes shows the image); every other file is offered as a file URL **only**, so it attaches as a file rather than inserting its contents (the `shouldInline` rule, applied identically on host and guest). The wire only ever carries the filename, never a host/guest path; the staged URL is synthesized locally after `sanitizedForApply`. Over-cap files are rejected/skipped rather than degraded to a name string. Host staging orphans are reclaimed at app launch (mirroring the agent's sweep-on-start).

## Window rebuild (`Views/Clipboard/`)

- Content area switches by `ClipboardPreviewPolicy`: a **file payload** shows as the file (image inline, any other file as a **chip** — type icon + name + "type · size"); otherwise an image preview (CGImageSource thumbnail capped at 2048 px so a 100-megapixel paste can't balloon memory), **styled RTF** (read-only; HTML is deliberately *not* rendered styled — `NSAttributedString`'s HTML import can synchronously fetch remote resources), an editable plain-text editor (text ≤ 2 MB), or a per-representation summary
- **Drag-and-drop works over the entire window** — the read-only previews no longer intercept drops (their image views unregister dragged types), and the editable editor diverts file/image/promise drags to the same intake path
- **Paste from Mac / Copy to Mac** buttons plus responder-chain `paste:`/`copy:` — the pasteboard is still only touched on explicit user gesture, preserving the gated-buffer posture; paste/drop send to the guest immediately, typed text still sends on blur
- Content-type indicator ("PNG image · 1920 × 1080 · 3.4 MB", "note.txt · Plain Text Document · 9 bytes") doubling as the transient surface for transfer failures — previously log-only
- The race-prone `isUpdatingFromService` flag is replaced by digest-based echo suppression (`ClipboardContent.digest`, SHA-256 over the (uti, data) pairs — filename excluded — computed once at init; dedup state holds 32 bytes, never duplicate multi-MB blobs)
- Undo/first-responder hygiene on mode switches; last-writer-wins conflict policy documented on the class

## Test plan

- [x] `make test` (all three targets), `make test-package`, `make lint` green locally — **1184 tests, 0 failures** (xcresult-confirmed)
- [x] New suites: `ClipboardContentTests` (digest canon, proto round-trip, policy table incl. receive-side caps, frame-budget sanity), `ClipboardFileStagingTests` (staging, supersede, path-traversal/dot-only sanitize), `ClipboardPreviewPolicyTests` (file/image/rich-text/text/summary priority), `ClipboardContentDescriberTests`, `ClipboardPasteboardIntakeTests`, `ClipboardContentAppKitTests` (inline rule + helpers); vsock service + guest agent suites extended (UTI flows, legacy interop both directions, file-only apply, over-cap skip, digest dedup, oversize/peer-error surfacing); `FakePasteboard` rewritten data-oriented
- [x] Manual end-to-end on the disposable macOS test VM (driven via computer-use): image + file copy/paste both directions — files materialize in Finder, images inline in Notes — RTF styled preview, and drag-over-image — **confirmed working**
- [x] `/review` (multi-agent, 5 dimensions × adversarial verification) run; 13 findings triaged and the in-scope ones fixed; #345 filed for an off-main-actor file-read follow-up


nicholas-lonsinger:
Two fixes from manual testing (1fcac62):

1. **Screenshot-thumbnail drag pasted a URL string** — the thumbnail drags a *file promise* (`NSFilePromiseReceiver`), whose only directly-readable representation is the URL text; the policy filters promise types, so the text won. The drop path now registers the promise drag types, receives the promised file into a scratch directory, and feeds it through the same file intake as a regular Finder drag (scratch copy removed after the bytes are buffered). Real file URLs and regular content still take the synchronous path.

2. **Pasting a large image ballooned the window** — `NSImageView`'s intrinsic size is the image size, and the preview's required constraints let it drive the window frame. The preview is now pinned to the container with floored compression-resistance/hugging priorities, so the *displayed* image aspect-fits whatever size the window has (buffered bytes untouched) — Quick Look-style fit rather than window growth or scrollers. Same priority floor applied to the summary view's labels. Note: a window frame that already autosaved huge needs one manual resize.

nicholas-lonsinger:
## Code review (self + independent adversarial pass)

**Verdict:** sound overall — protocol invariants, legacy interop, and the window's echo-suppression rework all held up under adversarial review. Four findings, all fixed in ee657b3.

### Findings → fixed

| # | Severity | Finding | Fix |
|---|----------|---------|-----|
| 1 | **Important (security)** | `ClipboardSnapshotPolicy` was sender-only: both `handleData` sites applied peer-supplied representations verbatim, so a malicious/buggy peer could hide a `public.file-url` rep behind a visible PNG; "Copy to Mac" would then plant the attacker-chosen path on the host pasteboard (confused-deputy across the VM→host boundary — a *new* exposure vs. the text-only pipeline) | `ClipboardSnapshotPolicy.sanitizedForApply` applied on receipt in both `VsockClipboardService.handleData` and `VsockGuestClipboardAgent.handleData`, with warning logs + tests on all three layers |
| 2 | Moderate | Promise receipt deleted the shared scratch directory per file (recursive `removeItem` could race a multi-file promise's pending writes), and the cleanup `defer` sat after `guard let self` — closing the window mid-receipt leaked the scratch dir | Per-file cleanup before the self guard; directory removed only once drained; first-file-wins gate matching the single-item buffer |
| 3 | Minor | Guest-pushed content between 12–16 MiB re-fired the "too large" transient on every window blur (`grabIfChanged` re-created the issue with a fresh `date`) | Oversize guard marks the content digest handled — one transient per distinct content |
| 4 | Minor | Unusable inbound payloads (all-forbidden reps, invalid UTF-8, unknown format) returned without consuming `pendingInboundGeneration` | Generation cleared on every unusable-payload path, both sides |

### Verified-fine under suspicion (no action)
- Generation/commit-after-send invariants survived the content-model rewrite (incl. the round-trip dedup reset); legacy↔new interop correct in all four handler×direction combinations, including image-only offers being invisible to legacy peers
- `lastAppliedDigest` echo suppression: no keystroke/guest-push ordering found that rebuilds the editor under the user; documented last-writer-wins holds
- MainActor bridging (promise completion `Task { @MainActor }`, transient-revert task, agent serve publish/cleanup) and closure capture lifetimes are clean
- Image preview can no longer drive window size (priorities floored, 2048 px decode cap)

### Dismissed (negligible by design)
- Bounded main-thread work: ≤10 MiB file reads in intake, ≤~12 MiB SHA-256 digests (~10 ms worst case)
- Multi-item drags/pastes take the first pasteboard item only — consistent with the single-item buffer model, documented
- Inbound rep-count flooding is bounded by the 16 MiB frame cap; receive-side size caps deliberately omitted (sender caps + frame limit are the contract); per-VM vsock auth hardening is already tracked in #145

nicholas-lonsinger:
## Follow-up: copied-file behavior now matches native macOS (d0fc4d9)

From testing — copying an image file in the guest only sent its filename, because a Finder ⌘C carries a *file reference*, not the pixels. Now a copied/dragged file transfers the way macOS treats it, symmetrically in both directions:

- **Image file → the image itself.** The guest agent (`expandedImageFileContent()`) reads the file's bytes when the pasteboard's first item is a single image file-url, so it pastes as an image on the host (matching "copy image file → paste into Notes shows the image"). The host window already did this for Paste/drag.
- **Any other file (.txt, .rtf, …) → just its name.** Contents are never inlined — it behaves like "the file," not its text. This also changed the host intake, which previously inlined text-file contents.

Agent version bumped 0.11.0 → 0.12.0 (update prompt will fire). Tests added on both sides; `make test` + `make lint` green. CI running.

Note: requires the new agent — copying an image file with the old 0.11.0 agent still only sends the name.

nicholas-lonsinger:
## Fix: dragging a screenshot now shows the image, not the path (1addc80)

Diagnosed via a multi-agent investigation (code trace + macOS drag-pasteboard research + canonical AppKit pattern, adversarially synthesized). **Confirmed root cause:** the floating screenshot **thumbnail** is a file-*promise* drag — no concrete `public.file-url`, no inline image bytes, only a `promised-file-url` (pointing at the temp file screencaptureui has already written) plus a path-text fallback. `NSFilePromiseReceiver` misses screencaptureui's legacy promise, so control fell to the generic snapshot, where the path text survived the skip-filter and then `ClipboardPreviewPolicy`'s text-wins rule showed it, hiding any image. Two compounding defects. (A desktop screenshot *file* drag already worked — it's a concrete file-url.)

**Fix (localized, no global filter widening):**
- `read(from:)` resolves a concrete-OR-promised file already on disk (`existingFileURL`) → reads the thumbnail's temp file as the image; and for any file/promise drag it drops the URL/path text fallbacks so a path can **never** leak as content.
- `handleDrop` is image-first: synchronous intake (inline image / on-disk file incl. the thumbnail / text), then async `NSFilePromiseReceiver` only if nothing resolved. Worst case is now a clean rejection, never a path.
- `ClipboardPreviewPolicy` lets an image beat a coexisting path/URL *descriptor* (Safari image drag) while genuine prose still wins (RTF+text).
- `shouldSkipBeforeReading` deliberately left as-is (a copied plain URL/text must still work).

**Verification:** deterministic unit tests reproduce the exact mechanism — `promisedFileURLImage` (promised-file-url on disk → image, **path does not leak**), `promiseWithoutFileNoPathLeak` (no on-disk file → rejection, no path), plus image-beats-URL-descriptor vs prose-stays-text. Full suite + lint green. The one piece left is a live confirmation that screencaptureui's `promised-file-url` resolves on a real thumbnail drag (the temp-file path in the original report indicates it does); a `.debug` "Clipboard drop types" log was added to read the exact types if a real drag ever still misses.

nicholas-lonsinger:
## ✅ Verified fixed in the real app (live drag + unified log)

The earlier intake/preview fixes were necessary but **not sufficient** — and the unified log revealed why: dragging onto the editor, my `handleDrop` was **never called**. The editable `NSTextView` is itself a drag destination and swallowed the drop, inserting the file's path as text before the window's image-aware intake ever ran. (No `Clipboard drop types` line on a real drag; a 135-byte path-text offer went to the guest.)

**Real fix (a08cb31):** `ClipboardEditorTextView` (NSTextView subclass) diverts file/image/file-promise drags to `handleDrop`, leaving plain-text drags and in-editor text moves alone.

**Ground truth from a live screenshot-thumbnail drag (now that `handleDrop` runs):**
```
Clipboard drop types: com.apple.NSFilePromiseItemMetaData, com.apple.pasteboard.promised-file-name,
  …promised-file-content-type, …NSFilePromiseID, public.file-url, public.png, dyn…, com.apple.pasteboard.promised-file-url
Took in pasteboard content (1 reps, 501299 bytes)   ← the PNG image, not a path
```
Before: 135-byte path text. After: a ~501 KB image taken and offered to the guest. User confirmed the window now shows the image.

So the full set of commits for this bug:
- d0fc4d9 / 1addc80 — image-aware intake (promised-file-url resolution, file-context text-fallback skip, image-beats-path preview) — correct, but unreachable through the text view.
- **26bd80a/a08cb31 — the decisive fix:** stop the editor from swallowing the drop.

All three CI checks green.

nicholas-lonsinger:
## ✅ Cross-boundary image-file paste verified live (both directions)

Live test on the macOS test VM (agent 0.13.0): copying an image file on one side and pasting it on the other **creates the file in Finder AND inlines in Notes**, both host→guest and guest→host. Host log corroborates (~500 KB images crossing): `Took in pasteboard content (1 reps, …)` + `Sent clipboard offer` (host→guest), `Received guest clipboard content (… bytes)` + `Copied clipboard buffer to host pasteboard` (guest→host).

**Mechanism (researched, 439d50c):** Finder "Paste" does not fulfill an `NSFilePromiseProvider` on the pasteboard — that's drag-only. So the receiver **materializes** the bytes to a local temp file (`ClipboardFileStaging`, one generation per write, superseded each apply, swept on teardown) and puts that file's `public.file-url` next to the inline image in one pasteboard item. The wire carries only the filename (additive `ClipboardRepresentation.filename`, excluded from the digest so echo suppression holds); the file-url is synthesized locally after `sanitizedForApply`, so no host path ever crosses. Agent bumped 0.12.0 → 0.13.0.

The design is already file-type-agnostic — only the capture side is gated to images. Generalizing to **any** file type (the north-star "anything copy/pasteable") is tracked as a follow-up.

PR #402, opened 2026-06-25
feat: Dissolve the 256 MiB inline clipboard size cap

### LINKED ISSUE #370: Image files over 256 MiB can't be pasted into the guest (force-inlined, rejected by maxInlineBytes)
## Found during
Root-causing a ~20 GB file paste that hung then failed silently (the absolute lazy-pull timeout fix). Surfaced by the adversarial review pass.

## Description
An image *file* on the clipboard is classified as **inline** by `shouldInlineOnPasteboard` / `VsockGuestClipboardAgent.shouldInline` (a file with a filename inlines iff its UTI conforms to `public.image`). Inline representations are reassembled in RAM on the receiver and capped at `ClipboardStreamTuning.maxInlineBytes` (256 MiB). The host sender sets `is_inline = true` for the image file, so the receiver's `ClipboardStreamReceiver.handleBegin` rejects any image file larger than 256 MiB up front with `inline.too.large` — before a single chunk streams.

Net effect: **any image file (PNG/JPEG/TIFF/HEIC/PSD/RAW/…) larger than 256 MiB cannot be pasted host↔guest at all** — it fails instantly. The disk-backed file path (which has no size cap) is bypassed purely because the file happens to be an image.

This is distinct from a non-image file of the same size, which streams to disk fine.

## Why it matters
Large images (camera RAW bursts, multi-layer PSDs, panoramas) are a realistic clipboard payload and silently fail. The inline classification exists so an image file can *also* be served as inline image data (e.g. paste-into-Preview), but it shouldn't *replace* the file path or impose the 256 MiB ceiling on the file itself.

## Location
- `ClipboardContent.Representation.shouldInlineOnPasteboard` (host) and `VsockGuestClipboardAgent.shouldInline` (guest) — the inline classification.
- `ClipboardStreamReceiver.handleBegin` — the `maxInlineBytes` guard that rejects it.
- `ClipboardStreamTuning.maxInlineBytes` — the 256 MiB cap.

## Suggested fix
Let an image *file* stream to disk as a file representation (uncapped) and additionally offer inline image bytes only when the payload fits `maxInlineBytes`; over the cap, fall back to file-only instead of rejecting. Design when picked up — the host/guest inline rule and the receiver's begin-time gate both need to agree.

### LINKED ISSUE #393: perf: Memory-map staged files when materializing inline pasteboard bytes
## Found during
Design/review of #370. Once an image *file*'s inline bytes are served from a staged file (rather than a RAM-reassembled buffer), the materialization path reads the whole file into a heap buffer.

## Description
When Kernova materializes inline pasteboard bytes from a staged file it reads the entire file into a contiguous heap buffer via `Data(contentsOf:)`. Switching to `Data(contentsOf:options: .mappedIfSafe)` maps the file instead: pages fault in on demand and are reclaimable (clean, file-backed), so there is no single allocation the size of the payload, and the handoff to the pasteboard server moves the pages copy-on-write. Disk I/O is deferred to actual page access. The larger the file, the larger the reduction in the CPU/RAM/throughput overhead Kernova adds on this path.

Today this affects the host read; once #370 serves an image file's inline bytes from disk on the guest, the guest gains the same read and should use the mapped form too.

## Why it matters
Lowers the processor/RAM/disk overhead Kernova adds to a copy/paste — especially for large files — by not allocating and eagerly reading the full payload just to hand it to the pasteboard. Better balances RAM against disk I/O on the inline-materialization path.

## Location
- Host: `ClipboardContentViewController.inlineData(for:)`.
- Guest: `VsockGuestClipboardAgent.provideData` (the inline-bytes branch introduced by #370).

## Considerations
- `.mappedIfSafe` maps only when the runtime deems it safe (volume/size) and otherwise reads normally — safe as a drop-in.
- A mapped `Data` stays coupled to the staged file for its lifetime (truncation/deletion while mapped would fault). Kernova owns the staging files and they are stable for the paste lifetime; the read should occur while the staged file is guaranteed present.
- Related: #370, #376, #371 (large-file paste disk usage).

nicholas-lonsinger:
## Summary

Pasting an image file larger than 256 MiB across the host↔guest boundary failed: image UTIs were force-inlined regardless of size, and the receiver hard-rejected inline payloads over the 256 MiB `maxInlineBytes` ceiling with `inline.too.large` — *after* the user had already clicked paste.

Per CLIPBOARD.md's triage (§1 no size bound, §2 disk-as-fallback), the fix is **not** "fall back to file-only over the cap" — it makes inline residency an implementation detail. A large inline representation now spills to the staging file (under the existing free-space guard, exactly like a file rep) and is served back via a memory-mapped read, so:

- inline content matches native capability at **any size** — no Kernova-imposed cap;
- the image still pastes **as image bytes** (full fidelity), never downgraded to a file-only paste;
- Kernova's added RAM stays near zero — the mapping pages in on demand and the OS can evict it.

This resolves #370 and #393 as a single task (CLIPBOARD.md folds #370 into #393).

Closes #370, closes #393

## Changes

- **`ClipboardStreamReceiver`** — `streamsToDisk` discriminator; an inline rep past `maxResidentInlineBytes` streams to the staging sink and is materialized as `.inMemory(Data(contentsOf:options:.mappedIfSafe))` at End. `inline.too.large` reject removed; the disk free-space guard now bounds inline payloads. Residency threshold is injectable for tests.
- **`ClipboardStreamTuning`** — `maxInlineBytes` → `maxResidentInlineBytes`, redocumented as a spill point, not a cap.
- **`VsockClipboardService`** — post-pull content digest computed off the main actor (`makeOffActor`) so an uncapped inline payload never stalls the owning actor (§8). The guest serves received bytes lazily and suppresses echoes by changeCount, so there is no guest-side hashing regression.
- **`ClipboardContentViewController.inlineData`** — memory-mapped read + symmetric host-side cap gate removed.
- **KernovaGuestAgent** → `0.27.0` (behavior change; wire format unchanged, so no capability gate — an un-updated guest simply still rejects large inline, no worse than today).

### Why `.inMemory(mapped)` and not a new disk-backed `Source` case

A spilled rep's content digest must stay consistent with the same bytes read back from the pasteboard (`computeDigest` hashes an `.inMemory` source's raw bytes, domain tag 0). A `.file`-style source would hash via its stored SHA-256 (tag 1) and diverge, **breaking echo suppression** (the guest would echo a huge image straight back). Keeping `.inMemory` preserves that consistency; the one off-actor full hash is the accepted price.

## Test plan

- [x] `make test-package` — 138 pass, including 3 new spill tests (spill→mmap round-trip with echo-suppression digest check, below-threshold stays resident, spilled partial cleaned up on cancel)
- [x] `make test` — `** TEST SUCCEEDED **`, full suite green (rebased onto #401)
- [x] `make lint` — clean
- [ ] End-to-end: install the 0.27.0 agent in the disposable test VM and paste a >256 MiB image both directions


nicholas-lonsinger:
Ran a multi-angle self-review (`/code-review` xhigh). One real finding, now fixed in 062260a: `republishOffActor` snapshotted the promise's materialized reps before its off-actor digest hash, so under concurrent preview + Copy-to-Mac materialization a stale snapshot's `apply()` could land after a fresher one and transiently revert a materialized rep to a placeholder chip (low-probability — a stale snapshot is a subset and normally hashes faster — and no paste-correctness impact, since Copy-to-Mac uses its own local result). Fixed with a monotonic `materializeEpoch` re-checked after the hop. Other candidates (e.g. `map.error` logged at `.debug`, `try?` on the mmap read, `.mappedIfSafe` fallback) were pre-existing patterns or not triggered in practice.

nicholas-lonsinger:
Rebased onto `main` (now at #401, which independently bumped the guest agent `0.25.0 → 0.26.0`). Two notes:

- **Guest agent advanced to `0.27.0`.** #401 took the same `0.26.0` slot this PR had picked, so per the version-bump rule the later PR advances to a unique version. The rebase surfaced this as the expected silent pbxproj non-conflict (git treated my `0.25→0.26` edit as already-applied), not a textual conflict.
- **One ARCHITECTURE.md conflict**, in the `KernovaGuestAgent` bullet — resolved by keeping #401's Status-submenu/symmetric-clipboard prose and grafting in the spill/mmap residency wording + the `0.27.0` version-history entry.

`make test` (`** TEST SUCCEEDED **`), `make lint`, and `make test-package` all green against the rebased base. No clipboard source files overlapped with #401.

PR #462, opened 2026-07-04
feat: MAS-compatible File Provider IPC + drop the LaunchAgent

### LINKED ISSUE #460: MAS-compatible File Provider: SMAppService.mainApp + NSFileProviderService + standard app group
# MAS-compatible File Provider: SMAppService.mainApp + NSFileProviderService + standard app group

Starting point for the full investigation — the target shape and the APIs involved, not a plan.
Based on patterns validated in the `backgroundtestapp` spike.

## What changes, and how it works

**A. Background registration → "Open at Login."**
`SMAppService.agent` (bundled LaunchAgent, "Allow in the Background") → `SMAppService.mainApp`
(the app itself opens at login). Uses `SMAppService` from **ServiceManagement**. `.mainApp`, not
`.loginItem`, because the VM runs in-process (`VZVirtualMachineView` must share the VM's process),
so the background process has to be the full app.

**B. File Provider host↔extension IPC → NSFileProviderService.**
Drop the LaunchAgent Mach-service. The extension conforms to `NSFileProviderServicing` and vends an
anonymous XPC endpoint (`NSFileProviderServiceSource.makeListenerEndpoint()` →
`NSXPCListener.anonymous()`). The host connects via `FileManager.getFileProviderServicesForItem(at:)`
→ `getFileProviderConnection` and passes an exported object, making the pipe bidirectional — so at
`fetchContents` the extension calls the host to fetch bytes (host stages into the shared container,
returns a path, extension clones). A Darwin notification (`CFNotificationCenterGetDarwinNotifyCenter`)
is the reconnect signal when the pipe is down (the extension can't initiate the connection itself).
Uses **FileProvider** (`NSFileProviderServicing`, `NSFileProviderServiceSource`,
`NSFileProviderManager`), **Foundation** (`NSXPCListener` / `NSXPCConnection`), and **CoreFoundation**
(Darwin notifications).

**C. App group → standard form.**
`8MT4P4GZL2.app.kernova` → `group.app.kernova` on the host targets — the team-ID prefix was only
needed for the Mach-service lookup, which goes away with (B).

Plus: the main app becomes sandboxed (it's currently un-sandboxed) for MAS.

## Next
Full investigation to scope the actual work.

nicholas-lonsinger:
Closes #460.

## What & why

Two coupled changes that make Kernova's File Provider stack MAS-forward and simpler:

**1. File Provider IPC → `NSFileProviderServicing` (host + guest).** The host↔extension byte relay moves off the launchd Mach service onto the canonical `NSFileProviderServicing` anonymous-XPC pattern. **Inverted wiring:** the owner (app / guest agent) is the XPC client and *exports* the relay; the extension is the server and calls it back at `fetchContents`. The owner sends an **`ownerDidConnect()` activation handshake** right after connecting so the extension's listener accepts the connection (an `NSXPCListener` fires `shouldAcceptNewConnection` only on the client's *first message*), and the byte pull is **non-blocking** so it can't deadlock the framework's serialised reconnect (which runs behind an in-flight `fetchContents`). A payload-free **Darwin notification** (`CFNotificationCenter`) is the reconnect doorbell; the vsock byte pull stays fully lazy.

**2. Drop the LaunchAgent → ordinary resident app + "Open at Login".** The launchd-managed `--background`/launcher/agent split is gone. The app is a plain resident menu-bar app; a **cold-launch activation heuristic** shows the window on a manual launch and stays headless at login (no public API distinguishes the two — Apple FB10207829). "Open at Login" is an opt-in `SMAppService.mainApp` **General-settings toggle** (default off).

Net: removes the launcher/agent roles, the `summon` XPC verb, the single-instance flock, and a KernovaKit protocol — shrinking the eventual App-Sandbox surface rather than growing it.

## Verification

- ✅ `make build` clean, `make lint --strict` clean, `make test` — **1514 passed, 0 failed**
- ✅ Multi-agent adversarial review across several rounds — every surfaced defect fixed (cold-launch `broadcast`; connect-at-registration; and for the servicing relay: the activation handshake, async pull, a `connectTimeout` data race, a fast-path retain-cycle leak, an accept-before-`resume` race, the doorbell re-probe, and retry-budget widening)
- ✅ **Real hardware (macOS TEST VM), guest→host "Copy to Mac":** 100 MB materialized in 0.7 s and **1 GB in ~6 s** — exact byte counts on disk, no error dialog. The extension logged `Accepted owner servicing connection` **for the first time**, and the connection accepted at VM resume stayed alive across pastes 40+ min apart (fast-path, no doorbell round-trips)
- ✅ Smoke test (launch model, computer-use): manual launch shows the window via the activation heuristic (`.accessory`→`.regular`); the **General** settings tab shows the Open-at-Login toggle; closing all windows leaves the app **resident + headless** (same PID); reopen returns the window
- ⏳ Not yet exercised: **host→guest** paste (uses the same fixed servicing code), and the login-launch heuristic via a logout/login cycle

## Note on the servicing relay (why it took a follow-up)

End-to-end hardware testing revealed the *initial* #460 servicing wiring **never delivered bytes** — guest→host "Copy to Mac" failed with *"Could not connect to Kernova."* Two root causes, both only reachable with a live guest↔host round trip (which is exactly why #460 deferred hardware testing):

1. **The extension never accepted the owner connection.** An `NSXPCListener` fires `shouldAcceptNewConnection` only on the client's first message, but the owner connected and exported its relay *without sending anything* — so the extension's listener never saw the connection and could never call the relay. → the `ownerDidConnect()` activation handshake.
2. **`fetchContents` deadlocked the reconnect.** It blocked waiting for the owner, but the framework serialises the owner's `getFileProviderConnection` / `supportedServiceSources` *behind* an in-flight `fetchContents`, so the owner couldn't connect until the 30 s timeout. → the non-blocking, doorbell-drained async pull.

Both fixed in `7fa33a3`; hardware-verified above.

## Deviations from the plan

- **Workstream 3 (app-group rename `→ group.app.kernova`) was dropped.** It broke the build (`"<target>" requires a provisioning profile`) — the team-ID prefix grants the app-group entitlement *implicitly* under profile-less manual signing ("Option A"), which the plan's premise ("team prefix was only for Mach lookup") missed. The team-prefixed group `8MT4P4GZL2.app.kernova` is already MAS-compatible; it's now only the shared staging container.
- **No programmatic legacy-agent unregistration.** With `app.kernova.plist` unbundled, `SMAppService.agent(plistName:)` resolves to `.notFound` and can't unregister the old registration — so a stale "Kernova" Login Items entry from an old dev build is removed by hand once (pre-release; no transition scaffolding).

## Follow-up: Finder cancel-button support for the fetch

Building on the async servicing pull, the extension now wires the `fetchContents` `Progress`'s `cancellationHandler`, so cancelling a "Copy to Mac" download in Finder completes the fetch promptly with `NSUserCancelledError` and stops the extension waiting — instead of ignoring the cancel until the pull finishes or times out.

- Each pull's `OnceCompletion` becomes the single arbiter across every terminal (owner reply, XPC error, reply timeout, connect timeout, **and cancel**), so a cancel racing a real completion can't double-fire the system completion handler. Cancel does **not** invalidate the shared long-lived servicing connection — it affects only the one pull.
- Adversarial review surfaced (and this fixes) a pre-existing `pull ↔ work-item` retain-cycle leak in `failPending`: the connect-timeout terminal never nil'd `connectTimeout`, unlike the fast-path / drain / cancel terminals.
- New `ClipboardFileProviderServiceSourceTests` cover the cancel path; `make test` is green.
- Aborting the **owner's** in-flight vsock transfer on cancel is out of scope here (bounded waste — the staged file is evicted by the owner's cache); tracked as #464.

## Canonicality review (design-level, vendor-doc grounded)

A three-track review verified the design against Apple's documentation and sample code:

- **The architecture is canonical.** The servicing pattern matches Apple's FruitBasket sample shape; the inverted owner-exports wiring is the DTS-documented "turn-around" anonymous-listener idiom (no first-party extension→app channel exists); the Darwin doorbell and non-blocking `fetchContents` match documented contracts; every alternative (launchd Mach service, CFMessagePort, app-group files, unix socket) is concretely worse under MAS constraints. `SMAppService.mainApp` + single resident binary is the canonical LaunchAgent replacement (a live `VZVirtualMachine` can't cross processes, so the agent form would duplicate the app).
- **One factual correction:** "no public API distinguishes a login launch from a manual launch" was wrong as stated — the launch `oapp` event's `keyAELaunchedAsLogInItem` property is public, legacy-documented, and field-proven against `SMAppService.mainApp` (Electron ≥ 29.4). The cold launch now resolves **deterministically** from the launch event (manual → window shows immediately, no 1.5 s settle; login item → headless), with the activation heuristic kept as fallback for an unreadable event.
- **One structural simplification:** the connect-timeout work item now weak-captures its pull, deleting the pull↔work-item retain cycle and the four-terminal manual cleanup discipline that had already produced one leak.
- The "framework serialises the owner's reconnect behind an in-flight `fetchContents`" rationale is now labeled as empirically observed (no Apple doc pins it); a stale `HostClipboardFileProvider` ARCHITECTURE.md entry was rewritten.
- Follow-ups filed: #465 (servicing state-machine test gap), #466 (connect-timeout ↔ retry-budget coupling).

## Guest agent

`MARKETING_VERSION` bumped `0.29.4 → 0.30.2` (File Provider servicing IPC changed, then the servicing relay was fixed, then the cancellation wiring + connect-timer refactor touched the agent-embedded extension — needs a reinstall).


nicholas-lonsinger:
## 🔍 Self-review (`/code-review`, xhigh) — findings fixed in 1303a88

A 10-angle finder → 1-vote verify → gap-sweep pass over the whole diff (26 agents). Everything below was verified and **fixed in this PR** unless noted; `make build`/`make lint --strict` clean, `make test` **1508 passed / 0 failed**.

### Correctness — File Provider servicing connector
| # | Finding | Fix |
|---|---------|-----|
| 1 | A reconnect doorbell coalesced while a connect is in flight was lost if that connect then failed, and a failed connect scheduled **no retry** → the extension's blocked `fetchContents` timed out (30 s) and the paste failed. | Bounded retry (`maxConnectAttempts`/`connectRetryDelay`) on every transient connect failure; documented that coalescing is now safe because the in-flight attempt retries. |
| 2 | `stopServing()` racing an in-flight connect could establish a **live relay-exporting connection after clipboard sharing was disabled** (relay read and connection stored under separate lock holds). | `configureAndResume` now re-checks `relayService` under the *same* lock that stores the connection; a concurrent disable drops it instead. |
| 3 | `handleConnectionDropped` cleared `self.connection` but never `invalidate()`-d it → on the interruption path (extension crash/relaunch) the connection↔handler retain cycle **leaked one NSXPCConnection per relaunch**. | `connection.invalidate()` on drop (no-op on the invalidation path, frees it on the interruption path). |

### Correctness — extension side
| # | Finding | Fix |
|---|---------|-----|
| 4 | `fetchStagedFile` waited on the reply semaphore with **no timeout** → a live-but-silent owner (stalled vsock pull) pinned a File Provider worker thread forever. | Bounded `semaphore.wait(timeout:)` (`fetchReplyTimeout`, off Finder's clock) → `serverUnreachable` on expiry. |
| 5 | `application(_:open:)` imported the VM but never summoned the window when the app was already **resident + headless** (the cold-launch heuristic had already latched, and no reopen fires for a document open). | Summon the GUI after import when `!isTestHost`; corrected the over-claiming `importVMs` doc. |

### Cleanup / tests
- Extracted the shared code-signing `anchor apple generic` + team-OU suffix into `ClipboardFileProviderConfig.teamSignedRequirement` so the two peer pins can't drift.
- Consolidated the duplicated `XCTestConfigurationFilePath` check into `ProcessInfo.isRunningXCTests` (shared by `AppDelegate` + `HostClipboardFileProvider`).
- Added `DarwinNotificationTests` (observer lifecycle / `cancel` idempotency / `deinit` teardown — the risky unretained-self code). The post→observe *delivery* round trip is intentionally not unit-tested: CFNotificationCenter needs a running main run loop the SwiftPM test target doesn't host, so it's left to integration.

### Deliberately not changed
- **Cold-launch settle timer** (`armColdLaunchActivationHeuristic`): a manual launch whose activation is delayed past 1.5 s latches headless. Annotated with a `RATIONALE` rather than restructured — no public API distinguishes a login from a manual launch (FB10207829), so a fixed window is inherent, and the always-present status item is the recovery path.
- **Cold-launch outcome enum**: kept — it's an intentional `classifyQuit`-style pure/testable seam, already covered by `AppDelegateColdLaunchTests`.
- **Extension replaced-connection leak**: REFUTED — the owner holds one connection at a time (strictly so after fix #3); a defensive `invalidate()` on replace was added anyway.

*Posting one summary rather than inline notes since every finding was fixed in the same pass — inline comments would be stale against 1303a88.*

nicholas-lonsinger:
## 🔬 Servicing relay fixed + verified on real hardware (`7fa33a3`)

The deferred end-to-end test surfaced that the #460 inverted-servicing relay **never actually delivered bytes** — guest→host "Copy to Mac" failed with *"Could not connect to Kernova."* Four real layers, each found from `os_log` traces on a live guest↔host round trip:

1. **Stale extension (environment).** The system was serving a pre-servicing `KernovaClipboardFileProvider.appex` from a ghost LaunchServices registration; reset it.
2. **Deadlock.** `fetchContents` *blocked* the extension waiting for the owner, but the framework serializes the owner's `getFileProviderConnection` / `supportedServiceSources` **behind** an in-flight `fetchContents` — so the owner couldn't connect until the 30s timeout. → made the byte pull **async** (return `Progress`, complete on connect/drain).
3. **Root cause — the connection was never accepted.** An `NSXPCListener` fires `shouldAcceptNewConnection` only on the client's **first message**, but the owner connected and exported its relay *without sending anything*. → added a **`ownerDidConnect()` activation handshake** (the "app calls the service" message servicing requires; confirmed vs. Apple's docs). This also makes the connection *durable*.
4. **Two adversarial-audit fixes** (2× 5-agent concurrency audits): `connectTimeout` data race, fast-path retain-cycle leak, and publish-before-`resume` on accept. Plus the doorbell now re-probes rather than tearing down a healthy connection, and the connect-retry budget spans the extension's 30s wait.

### Verified — `macOS TEST` VM, guest→host "Copy to Mac"
| File | `fetchContents` → materialized | On-disk |
|------|-------------------------------|---------|
| 100 MB | 0.7 s | 104,857,600 bytes (exact) |
| 1 GB | ~6 s | 1,073,741,824 bytes (exact) |

The extension logged `Accepted owner servicing connection` **for the first time**, and the connection accepted at resume stayed alive across pastes 40+ min apart (no doorbell round-trips). `make build`/`lint --strict` clean, `make test` **1514 passed / 0 failed**.

Guest agent bumped `0.30.0 → 0.30.1` (its File Provider shares this code). **host→guest** paste uses the same fixed path but wasn't exercised here — worth a follow-up check.

nicholas-lonsinger (commented):
[Superseded by review comment below]

nicholas-lonsinger (commented):
### PR Review Findings & Suggested Fixes

#### 1. Asynchronous `fetchContents` Refactor (Approved)
The asynchronous refactor of `fetchContents` using the `NSFileProviderServicing` anonymous-XPC connection is correct. Moving away from the synchronous thread-blocking `DispatchSemaphore` resolves the potential deadlock during reconnection.

#### 2. Timing Race: Placeholder Not Found on Immediate Paste (Suggested Fix)
Because `forceRootEnumeration` lists the directory asynchronously in the background, a user immediately pasting after a copy-click can cause Finder to query the File Provider `.fileURL` before the placeholder file is actually generated on disk, leading to `ENOENT` (file not found).

**Fix**: Resolve the `.fileURL` lazily inside `hostPasteboardItems` in `ClipboardContentViewController.swift`. If the File Provider placeholder does not exist on disk yet, fall back to pulling the file over vsock and staging it synchronously (exactly like the File-Provider-off fallback).

Update the `.fileURL` resolution inside `hostPasteboardItems` as follows:

```swift
// In ClipboardContentViewController.swift

                var types: [NSPasteboard.PasteboardType] = []
                let imageType = item.types.first { !$0.isFileURL }
                    .map { NSPasteboard.PasteboardType($0.uti) }
                if let imageType { types.append(imageType) }
                types.append(.fileURL)

                specs.append(
                    PasteboardItemSpec(types: types) { type in
                        if type == .fileURL {
                            // 1. If File Provider is enabled, check if the placeholder exists on disk
                            if let fpURL = representation.fileURL {
                                if FileManager.default.fileExists(atPath: fpURL.path) {
                                    return Data(fpURL.absoluteString.utf8)
                                }
                                
                                // 2. Fallback: placeholder is missing (timing race). Pull over vsock synchronously
                                if let fileProvider = service as? any HostClipboardFileRepProviding,
                                   case .success(let path) = fileProvider.pullStagedFile(
                                        generation: generation, repIndex: item.types[0].representationIndex) {
                                    return Data(URL(fileURLWithPath: path).absoluteString.utf8)
                                }
                            }
                            
                            // 3. Eager/Image fallback
                            if let stagedURL = stagedFileURL(
                                for: representation, generation: generation, staging: staging) {
                                return Data(stagedURL.absoluteString.utf8)
                            }
                            return nil
                        }
                        
                        if type == imageType {
                            if let stagedURL = stagedFileURL(
                                for: representation, generation: generation, staging: staging),
                               let data = try? Data(contentsOf: stagedURL, options: .mappedIfSafe) {
                                return data
                            }
                            return inlineData(for: representation)
                        }
                        return nil
                    })
```

nicholas-lonsinger:
### ⚠️ Design / Performance: Synchronous thread blocking in `fetchContents` using `DispatchSemaphore`
    
    #### **Problem**
    In ClipboardFileProviderExtension.swift, `fetchContents(for:version:request:completionHandler:)` uses a `DispatchSemaphore` to synchronously block the worker thread while the XPC relay pulls the file
over vsock and stages
  it.
    
    #### **Why this is non-canonical & problematic**
    1. **Defeats Progress & Cancellation**: `fetchContents` is designed to return an `NSProgress` object immediately so that the OS can monitor the download or cancel it. Because the thread is blocked,
  the `Progress` object is not returned to the system until the download is *already completed*. This prevents Finder from displaying a progressive download status or letting the user cancel a slow/stuck
  transfer.
    2. **Resource Starvation / Thread Exhaustion**: File Provider extensions run inside sandboxed worker processes. Blocking worker threads with semaphores under concurrent paste operations can lead to
  worker thread pool starvation.
    
    #### **Suggested Fix**
    Refactor `fetchContents` to run fully asynchronously. Return the `Progress` object immediately and manage the copy logic, completion handler invocation, and XPC connection lifecycle inside the XPC
  callback block:
    
    ```swift
        open func fetchContents(
            for itemIdentifier: NSFileProviderItemIdentifier,
            version requestedVersion: NSFileProviderItemVersion?,
            request: NSFileProviderRequest,
            completionHandler: @escaping (URL?, NSFileProviderItem?, Error?) -> Void
        ) -> Progress {
            logger.debug("fetchContents START (item=\(itemIdentifier.rawValue, privacy: .public))")
            let progress = Progress(totalUnitCount: 100)
    
            guard let decoded = ClipboardFileProviderItemIdentifier.decode(itemIdentifier.rawValue),
                let manifestItem = store.readManifest().item(for: itemIdentifier.rawValue)
            else {
                completionHandler(nil, nil, NSFileProviderError(.noSuchItem))
                return progress
            }
    
            let connection = NSXPCConnection(machServiceName: config.machServiceName, options: [])
            connection.remoteObjectInterface = NSXPCInterface(with: ClipboardFileProviderRelay.self)
            if let requirement = config.relayCodeSigningRequirement {
                connection.setCodeSigningRequirement(requirement)
            }
            connection.resume()

            // Wire up progress cancellation to invalidate the connection early
            progress.cancellationHandler = { [weak connection] in
                connection?.invalidate()
            }

            let proxy = connection.remoteObjectProxyWithErrorHandler { error in
                connection.invalidate()
                completionHandler(nil, nil, error)
            } as? ClipboardFileProviderRelay

            guard let proxy else {
                connection.invalidate()
                completionHandler(nil, nil, NSFileProviderError(.serverUnreachable))
                return progress
            }

            proxy.fetchFile(generation: decoded.generation, repIndex: decoded.repIndex) { [weak self] path, error in
                defer { connection.invalidate() }
                guard let self else {
                    completionHandler(nil, nil, NSFileProviderError(.noSuchItem))
                    return
                }
                guard let stagedPath = path else {
                    self.logger.error(
                        "fetchContents relay failed: \(error?.localizedDescription ?? "unknown", privacy: .public)"
                    )
                    completionHandler(nil, nil, error ?? NSFileProviderError(.serverUnreachable))
                    return
                }

                guard let manager = NSFileProviderManager(for: self.domain) else {
                    completionHandler(nil, nil, NSFileProviderError(.providerNotFound))
                    return
                }
                do {
                    let tempDir = try manager.temporaryDirectoryURL()
                    let destination = tempDir.appendingPathComponent(UUID().uuidString)
                    try FileManager.default.copyItem(at: URL(fileURLWithPath: stagedPath), to: destination)
                    self.logger.notice("fetchContents materialized \(manifestItem.byteCount, privacy: .public) bytes")
                    progress.completedUnitCount = 100
                    completionHandler(destination, ClipboardFileItem(manifestItem: manifestItem), nil)
                } catch {
                    self.logger.error("fetchContents clone failed: \(error.localizedDescription, privacy: .public)")
                    completionHandler(nil, nil, error)
                }
            }
            return progress
        }

PR #509, opened 2026-07-10
fix: Re-publish live inbound clipboard offer when File Provider domain becomes ready (#429)

### LINKED ISSUE #429: First host offer before the FP availability probe settles misses the File Provider path
## Found during
Self-review of PR #425 (#376 D1a guest File Provider).

## Description
Domain registration + the availability probe (`registerDomain` → `addDomain` → `refreshAvailability`) are asynchronous, so `availabilityStorage` is briefly not `.ready` after clipboard sharing is enabled — and, on a fresh VM, it sits at `.needsEnabling` until the user flips the System-Settings toggle. A large-file offer that arrives in that window hits `publishSingleFile`'s `availabilityStorage == .ready` guard, returns nil, and is written on the synchronous (60s-deadline-prone) provider path. When availability later flips to `.ready`, the agent does NOT re-evaluate the already-live `inboundPromise`, so the user must re-copy on the host to get an FP-backed placeholder for that same content. Steady state is correct; this is a startup / just-enabled-the-toggle rough edge.

## Suggested fix
When availability transitions to `.ready` in `refreshAvailability` (covering `.inactive→.ready` and the fresh-VM `.needsEnabling→.ready` flip), notify the agent (the host↔agent wiring is already bidirectional) to re-run its inbound FP routing for the LIVE `inboundPromise` — re-running `handleOffer`'s `fileProviderURLs`/`publishSingleFile` and re-writing the pasteboard item, preserving echo-suppression (`lastPasteboardChangeCount`), only when the current promise is the qualifying single non-inline/non-directory file rep. Keep the host↔agent isolation boundary main-queue-safe.

---
**Phase (clipboard arc, set 2026-06-28):** D1c hardening — re-publish the live offer when availability flips `.ready`.

nicholas-lonsinger:
## Summary
- A host offer for a single large file that arrives before the guest's File Provider domain settles (fresh VM, or the brief window right after enabling clipboard sharing) landed on the synchronous, 60s-deadline-prone provider path, and was never re-routed once the domain became ready — the user had to re-copy on the host to get an FP-backed placeholder.
- The guest agent now re-publishes the live inbound offer through the File Provider as soon as availability flips to `.ready`, so no re-copy is needed.

## Changes
- `VsockGuestClipboardAgent`: extracted the pasteboard-write logic from `handleOffer` into `writePasteboardPromise` (shared by the initial offer and the new re-publish path); added `InboundPromise.fileProviderRepIndex` to track whether the live promise is already FP-routed; added `fileProviderAvailabilityChanged(_:)`, which re-runs FP routing for the live promise only when it's still on the sync path, still the eligible single-file case, and the guest clipboard hasn't changed since it was written.
- `AgentAppDelegate`: wired the (previously unused, on the guest side) `FileProviderDomainHost` availability observer to the new method.
- Bumped the guest agent's `MARKETING_VERSION` (0.33.0 → 0.34.0) since this changes agent behavior and needs a reinstall to exercise.
- ARCHITECTURE.md: noted the new re-publish path and its test coverage.

## Test plan
- [x] `make build`
- [x] `make lint`
- [x] `make test-suite SUITE=KernovaMacOSAgentTests/VsockGuestClipboardAgentTests` (7 new tests covering the re-publish and its no-op guards)
- [x] `make test` (full suite, 1469 test cases, TEST SUCCEEDED)

Closes #429

https://claude.ai/code/session_01FiMiwLRkwTVXV7sg8DnzBZ


nicholas-lonsinger:
## Code review (medium effort)

Ran 8 parallel finder angles (line-scan, removed-behavior, cross-file trace, reuse, simplification, efficiency, altitude, conventions) against the diff, then verified candidates directly against the current source.

### Fixed in 6b6d8b4
- **Missing failure-path logging**: `fileProviderAvailabilityChanged`'s pasteboard-write failure branch silently dropped the promise with no `.warning` log, unlike `handleOffer`'s identical branch.
- **Duplicated publish/rollback bookkeeping** between `handleOffer` and `fileProviderAvailabilityChanged` — consolidated into a shared `publishPromise` helper.
- **Repeat log spam**: a live multi-file (File-Provider-ineligible) offer re-logged the "not routed through the File Provider" notice on every `.ready` transition instead of once.

### Filed as follow-up (out of scope for this PR)
- #510 — the availability-flip re-publish reuses the same `generation`/`repIndex` as the original sync-path routing, so a stale externally-retained pasteboard-item reference could in theory collide transfer IDs with a fresh `fetchContents` pull. Narrow window (requires an external reader to retain a not-fully-consumed promise across the switch), not blocking, but a real gap in the "never both paths" invariant `fetchStagedFile` documents.
- #511 — the host-side "Copy to Mac" outbound direction has the identical pre-#429 availability race (`FileProviderDomainHost.publishSingleFile` checks a stale `availabilityStorage` right after firing the refresh that would update it) — this PR only fixes the guest-inbound direction per #429's scope.

### Also considered, not changed
- Full pasteboard rewrite on re-publish bumps `NSPasteboard.changeCount` for co-offered unrelated items (e.g. an inline text rep alongside the file rep) — no cheaper `NSPasteboard` API exists to patch one item in place, so left as-is.
- Redundant `refreshAvailability()` re-probe inside `publishSingleFile` when called from a path that just confirmed `.ready` — real but minor inefficiency in shared `KernovaKit` code with two consumers; not worth the risk/diff size for this PR.

All 57 `VsockGuestClipboardAgentTests` and the full `make test` suite pass after the fixes.

Issue #603, opened 2026-07-18
Clipboard paste does not carry extended attributes (xattr) on any path


nicholas-lonsinger:
Filed as a follow-up from the folder placeholder-tree work (D1b folders, #422 / #376 arc), originally scoped to the placeholder-tree path only. Scope corrected 2026-07-20: the gap is universal across every paste path (details below).

## Observed
No clipboard paste path reads, transmits, or applies extended attributes. A pasted file or folder carries its data, structure, POSIX permissions, and mtime; xattrs (Finder tags, `com.apple.quarantine`, `kMDItemWhereFroms`, custom app attributes, …) never cross. Verified per path:

- **Placeholder-tree path** (`clipboard.dirtree.v1`): `ClipboardTreeEntry` / `ClipboardDirectoryTree.enumerateTree` record kind, size, permissions, mtime, symlink target, and package flag only — no xattr field exists on the wire — and no `NSFileProviderItem` conformer implements `extendedAttributes`.
- **Single-file paths** (File Provider placeholder and sync stream): file bytes are streamed and written into a freshly created destination file (`ClipboardStreamReceiver`); xattrs are never read at the source. The only `copyItem` on the paste path copies from that fresh staged file.
- **Eager-archive fallback**: contrary to this issue's original premise, `ClipboardDirectoryArchive.fieldKeys` is `TYP,PAT,LNK,DEV,DAT,UID,GID,MOD,FLG,MTM,CTM,SH2` — no `XAT` — so AppleArchive does not preserve xattrs either, and `ClipboardDirectoryArchiveTests` has no xattr assertion.
- **Repo-wide**: zero uses of `getxattr`/`listxattr`/`setxattr`/`extendedAttributes` in any target.

## Impact
Uniform across all paths and both directions: pasted copies lose Finder tags, quarantine, "where from", and custom app xattrs. Because the behavior is identical everywhere, the user-visible result is consistent (if lossy).

## Research (sourced, 2026-07-20)
`NSFileProviderItem.extendedAttributes` does stamp real on-disk xattrs that survive Finder copy-out — but only for names flagged `XATTR_FLAG_SYNCABLE` (`#S`), under a ~32 KiB per-item system budget (larger attributes such as resource forks are stripped regardless). The placeholder-tree path's ceiling is therefore small-xattrs-only even if wired.

## Resolution
[...]

PR #602, opened 2026-07-18
feat: Determinate progress for File Provider clipboard pastes (#426)

nicholas-lonsinger:
## Summary
The File Provider large-file paste (#376) now shows a **determinate** download bar in Finder — in both directions — instead of the pulsing "Preparing to copy…" indeterminate bar, and the host clipboard window's in-app bar reflects the guest→host File-Provider-triggered pull. PR 3 of the clipboard File Provider arc.

**No vsock wire change** — progress travels over the *existing* `NSFileProviderServicing` XPC channel, not the transport. (Ground-truth correction from the design pass: the outbound in-app half already recorded via `handleRequest`; the real in-app gap was only the host receive side, `performBlockingPull`.)

## Changes
- **XPC contract** (`FileProviderRelay.swift`): one-way `fetchProgressed(generation:repIndex:bytesTransferred:totalBytes:)` on `FileProviderControl` — the owner pushes coalesced per-chunk bytes back to the extension.
- **Extension** (`FileProviderServiceSource.swift`): in-flight progress-handler registry keyed by `(generation, repIndex)` (token-guarded against the #500 same-key retry), registered/unregistered over `fetchStagedFile`'s lifecycle; a late push for a finished pull or an unknown key no-ops.
- **`fetchContents`** (`FileProviderExtension.swift`): byte-denominated `Progress` (`totalUnitCount = manifest byteCount`, zero-guarded; `kind = .file`, `fileOperationKind = .downloading`), advanced from pushes and completed in `materialize`. Both appex subclasses inherit.
- **Owner push** (`FileProviderRelayService.fetchFile`): captures the calling `NSXPCConnection` synchronously and drives a throttled `FetchProgressPusher` (pure `FetchProgressThrottle` — ~1% of total or ≥100 ms apart, always the final chunk) fed by `FileProviderPullProvider.fetchStagedFile`'s new `onProgress`. A version-skewed peer lacking the selector degrades to no-progress without tearing down the connection.
- **Pull chain**: `onProgress` threaded through `HostClipboardPullRouter`/`HostClipboardFileRepProviding.pullStagedFile` + `VsockClipboardService.performBlockingPull` (host) and `VsockGuestClipboardAgent.fetchStagedFile`/`pullRepresentation` FP-relay path (guest); the existing heartbeat re-arm is preserved.
[...]

Issue #499, opened 2026-07-09
Give clipboard transferIDs a per-attempt discriminator to close a stale-cancel/abort collision window


nicholas-lonsinger:
## Found during
Review of PR #495 (#464 — propagate File Provider fetch cancellation to the owner's vsock pull).

## Description
`ClipboardTransferID.make(generation:repIndex:hostMinted:)` derives a transfer id purely from `(generation, repIndex, direction)` — the same still-current offer pulled twice (a retry, or a second read) always produces the identical id, with no per-attempt nonce/epoch.

Both `ClipboardStreamReceiver.cancel(transferID:)` and `ClipboardStreamSender.handleAbort(transferID:)` address in-flight state purely by this bare id. If pull attempt #1 for `(G, R)` is cancelled and its cancel (a `ClipboardStreamAbort` wire frame, or a local `cancelStagedPull` call) is delayed just as a second, legitimate pull attempt for the identical `(G, R)` starts and registers under the same id, the stale cancel can land on the new attempt instead — aborting a transfer that should have continued.

Verified during review: the mechanism is real (no discriminator, both sides key purely on the bare id), but realistically triggering it requires a sub-millisecond re-pull of the exact same offer/rep racing an in-flight cancel plus adversarial scheduling — not something normal Finder/File-Provider behavior produces (a re-paste happens seconds after a cancel completes, long after the stale cancel has already settled to a no-op).

## Location
`ClipboardTransferID.make(generation:repIndex:hostMinted:)`, `ClipboardStreamReceiver.cancel(transferID:)`, `ClipboardStreamSender.handleAbort(transferID:)`.

## Suggested fix
Mint transfer ids with a per-attempt epoch (e.g. a monotonically-incrementing counter folded into the reserved high bits, or keyed off `LazyPullCoordinator`/`ClipboardStreamReceiver` awaiter-registration generation) so a stale cancel can't address a later incarnation of the same `(generation, repIndex)`. Design when picked up.

Issue #376, opened 2026-06-22
feat: Use a File Provider domain for large-file clipboard paste (on-demand materialization, no 60s Finder deadline)


nicholas-lonsinger:
## Summary
Replace the pasteboard **promise** the guest puts on the clipboard for an inbound file with a **concrete file URL inside a File Provider domain** (`NSFileProviderReplicatedExtension`). A paste then returns instantly and the bytes are materialized **on read** via `fetchContents(for:request:completionHandler:)`, with native macOS progress and no time limit.

## Motivation
Finder abandons a pasteboard promise after **exactly 60 seconds** (measured on macOS 26.5.1; logged by Finder as `Finished waiting for promise because timed out`), independent of and stricter than the 120 s generic `NSPasteboard dataForType:` floor. The guest agent's synchronous `provideData` (`LazyClipboardDataProvider`) blocks while materializing the file, so any clipboard file needing >60 s to stream into the guest **fails the first paste** — the user sees a 60 s beachball, then nothing pastes; a second paste works only because the completed transfer is cached. This is **not** fixable by lengthening Kernova's own lazy-pull backstop (PR #372 already made it a re-arming inactivity window): the `NSPasteboardItemDataProvider` API has no "still working" signal to re-arm Finder's 60 s clock.

File Provider is the only public API that makes large-file paste succeed first-try at any size: it puts a real (dataless-placeholder) file URL on the pasteboard, so there is **no promise for Finder to time out**, and it defers the byte transfer to the file *read*, which has no 60 s cap and gets native progress UI. It is the same mechanism iCloud Drive / cloud-storage apps use, and is sandbox / Mac App Store-friendly.

## Scope
- A File Provider extension + a registered domain (`NSFileProviderManager`) representing files pasted from the host.
- On an inbound clipboard offer, present the file(s) as placeholder item(s) in the domain and put the domain file URL(s) on the guest pasteboard instead of a `public.file-url` promise served by `LazyClipboardDataProvider`.
- `fetchContents` pulls the representation from the host over the existing chunked vsock stream and writes it to the system-provided URL, reusing `ClipboardFileStaging` / the streaming protocol and free-space guard. The extension is sandboxed and likely cannot open a vsock directly — relay the pull through the guest agent (IPC).

## Phasing (one logical item, phased)
- **Phase 1:** Single-file paste end-to-end through the domain: placeholder → `fetchContents` → host pull → file appears with native progress, no 60 s limit.
[...]

PR #604, opened 2026-07-18
feat: Folder placeholder-tree paste with per-child fetch (D1b folders)

### LINKED ISSUE #422: Make directory clipboard transfer lazy (drop eager source archiving; consider streaming extraction to avoid the intermediate .aar)
## Found during
Planning PR D1 (#376, guest File Provider transport) — reviewing how folders cross the clipboard boundary.

## Description
A folder (or OS package such as `.app`/`.rtfd`) currently crosses the clipboard as a single AppleArchive (`.aar`, LZFSE) directory representation, **staged on disk at both ends**. The approach is correct and has no capability ceiling, but its disk footprint is a serious violation of CLIPBOARD.md §2 (no reflexive intermediate copy), §3 (pay on consume), and the resource ordering (peak must stay bounded as payload grows).

### Worst-case end-to-end disk footprint
For a host→guest paste of a directory of size **N** (incompressible, so `.aar` ≈ N), at peak Kernova holds, simultaneously, **on top of** the user's destination copy:

- **Source `.aar` (the copy-from side): ≈ N.** Built *eagerly at copy time* (the offer needs the archive's exact `byteCount` + SHA up front) and kept under the source's staging generation — a standing on-disk duplicate of the folder for the whole offer lifetime, **whether or not the user ever pastes**.
- **Received `.aar` (the paste-to side): ≈ N.** Streamed to disk in full and SHA-verified before extraction.
- **Extracted tree (the paste-to side): ≈ N.** Written next to its own `.aar` (the archive is not deleted after extraction — it lingers until generation sweep).

So Kernova's *own* marginal disk peaks at **≈ 3N** — a full copy of the payload on the source, plus a full copy *and* its archive on the receiver — before the legitimate destination write (a 4th N). For a large directory this is enormous and scales linearly with payload, which the ordering forbids. Prime "smarter software earns its keep" case (§12): the bytes should flow source-tree → wire → destination-tree with Kernova's added disk approaching zero.

### Two root costs (one cause: a directory crosses as a fully materialized archive)
1. **Eager archiving on the copy-from side (§3).** Archiving happens at copy, not at paste — a standing full-size duplicate of the source folder, paid even when nothing consumes it. (`ClipboardDirectoryArchive.archivedRepresentation(ofDirectoryAt:...)`; the host intake and the guest's `archiveAndOffer` outbound path.)
2. **Intermediate on-disk `.aar` on the paste-to side (§2).** The received archive is staged whole, then extracted, so the compressed archive and the uncompressed tree coexist. Extraction is already streaming/low-RAM but reads from the on-disk `.aar` rather than the live transfer. (`ClipboardDirectoryArchive.extractedDirectoryURL(...)` / `extract(archiveAt:to:)`.)

## Why it matters
Everything else in the subsystem is lazy and streamed; folder transfer is the outlier, and its peak disk scales as a multiple of payload size on **both** sides. Dissolving it brings directories in line with §2/§3 and the ordering, and removes the only place Kernova's own disk footprint grows with the payload.

## Considerations
- Lazy source archiving means the offer protocol must stop requiring an exact `byteCount` + SHA up front for directories — a negotiated capability (size estimate + post-hoc verification, or defer archiving to first paste), gated by `min_required_agent_version` (no legacy fallback).
- A zero-intermediate path means streaming the archive on the send side and streaming extraction on the receive side (decode straight into the tree without landing the full `.aar`), with the SHA-256 (the only corruption check, §7) computed **inline** during the stream and a rollback (delete the partial tree) on mismatch — trading "verify-then-extract" for "verify-while-extract."
- File Provider folder handling (D1b) may represent a directory as enumerated child items rather than a single archive fetch — design this *after* D1/D2 so it builds on the final materialization path.

## Sequencing
Do this **after** the File Provider transport arc (D1 #376 + the D2 host "Copy to Mac" mirror). Independent of the 60s-deadline fix, and it interacts with the materialization path D1/D2 reworks. Design when picked up — no implementation sketch here intentionally.

---
**Phase (clipboard arc, set 2026-06-28):** after D1/D2 — interacts with the final materialization path the File Provider transport reworks; design against it, not before.

nicholas-lonsinger:
## Summary
Final PR of the clipboard File Provider arc: a copied **folder** pastes as a File
Provider placeholder *tree* with per-child fetch — no archive built at copy time on
either side, so Kernova's marginal disk ≈ 0 and paste serves *current* file
contents exactly like a native Finder folder copy (decision 3 of the design of
record, #376). Fidelity moves from AppleArchive to File Provider item metadata.

Closes #422.

(Note: #376 is closed after the live-verification session, not by this PR body.)

## Wire protocol
- **Capability** `clipboard.dirtree.v1`, mutually negotiated on the control-channel
  Hello. A directory rep crosses as a placeholder tree only when **both** sides
  advertise it; otherwise it falls back to the eager-archive path (unchanged for
  old peers). No offer-shape field — the negotiated bool determines interpretation
  on both ends, and a capability mismatch degrades gracefully to the archive path.
- **Offer**: with the capability, a directory rep carries a stat-walk **size
  estimate** and no archive (the producer retains the source folder); without it,
  today's exact-archive offer.
- **`ClipboardTreeFetch`** (one new frame): empty `relative_path` = the tree
  **listing** (producer walks the source, streams a serialized `ClipboardTreeListing`
  inline over the existing SHA-verified transport); non-empty = one **child file**
  (confined resolution, streamed as a file). Replies reuse Begin/Chunk/End/Ack/Abort.
- **Child transfer-ID scheme**: a second high bit (62) selects a child layout
  `[gen:24][repIndex:16][childSeq:22]` (childSeq 0 = listing, ≥1 = a tree node),
  preserving the direction bit and the `(gen, repIndex, childSeq)` determinism the
  cancel/abort routing (#499/#500) relies on; `generation(of:)` is layout-aware so
  supersession cancel still matches child transfers.

## File Provider tree
`FileProviderManifest` is hierarchical (flat `items` + `folders`/`FolderRep`/`Node`);
`clipnode` tree-node identifiers; the enumerator serves folder roots + children; a
child `fetchContents` relays a child-addressed pull (with #602 progress). Paste-time
routing fetches the listing (bounded, in the barrier budget), publishes the root +
tree as dataless placeholders, returns the folder URL. Directory reps join the
FP-eligible set on **both** sides when the capability is negotiated.

## Fidelity (§6)
Symlinks (`symlinkTargetPath` + `UTType.symbolicLink`), executable/permission bits
(`fileSystemFlags`), empty directories, modification dates, and packages (folder
`contentType`, so a pasted `.app` is a package). **Extended attributes are not
carried through the FP item surface** — documented in CLIPBOARD.md §6 and filed as
#603 (verify live before relying on xattr round-trip); the archive fallback keeps
the full AppleArchive key set.

## Fallback
Capability absent or FP toggle off: the archive route is kept but the archive is
built at **request** time, not copy time. The deadline-safe total gate
(≤`maxDeadlineSafeFileBytes`) bounds the fallback's intermediate — the rationale
that lets #422 close.

## Live-verification checklist
- [ ] host→guest folder paste — placeholder tree, per-child determinate progress, no size cap
- [ ] guest→host folder "Copy to Mac"
- [ ] `.app` bundle paste with intact code signature (fidelity acid test)
- [ ] symlinks / executable bits / empty directories preserved
- [ ] toggle-off capped archive fallback still works (both directions)
- [ ] old-agent interop (capability absent → archive shape)

## Test plan
- [x] `make build` green
- [x] `make test` green (1717 passed / 0 failed)
- [x] `make lint` green
- [ ] Live verification per the checklist above


nicholas-lonsinger:
## Live VM verification — PASS (with two fixes landed during verification)

Verified on a freshly created macOS TEST VM (clean install, agent installed from scratch), host running this branch's staged build, agent 0.45.1, `clipboard.dirtree.v1` negotiated both ways.

### Defects found live and fixed on this branch

1. **Package containers broke every `.app`/`.rtfd` paste with Finder error -36** (4fdaaed). A package-conforming `contentType` on a placeholder-tree container makes the system fetch the container as one atomic file instead of enumerating children (`fetchContents` observed on the `childSeq 0` root via guest `log stream`). Containers are now always served as plain `.folder`; the pasted copy's packageness comes from its on-disk extension/bundle bit.
2. **Pasted folder roots landed with an epoch date (Dec 31 1969)** (4fdaaed). The listing carried descendant mtimes but not the root's own. `ClipboardTreeListing.root_mtime_ms` added; producer stats the source root; both consumers thread it through.

### Checklist results

- ✅ **Old-agent interop** (agent 0.44.0, capability absent): folder crossed via the eager-archive offer shape (exact archive byte count in the offer) and pasted correctly.
- ✅ **h→g folder placeholder tree**: 6 GB tree (nested dirs, 6 GB file, symlink, executable, empty dir) — offer carried the stat-walk estimate instantly (no copy-time archive), pasted with per-child determinate Finder progress. Verified in-guest: SHA-256s match the host, symlink target intact, exec bit preserved, empty dir present, per-child **and root** mtimes preserved.
- ✅ **`.app` signature acid test (h→g)**: Calculator.app materialized as an Application with working icon; all 20 files byte-identical to the source (matching sorted digest-of-digests). `codesign --verify --deep --strict` parity: the pasted copy behaves exactly like a local `cp -R` of the same sealed system app (both report the trust-cache-related "resource envelope is obsolete"; the source verifies via the trust cache).
- ✅ **Positive signature control (g→h)**: the guest's dev-signed `Kernova Guest Agent.app` copied to the host via the placeholder tree (host log: `FP published 0 file(s) + 1 folder(s)`, publish barrier verified, routed at paste) and passes `codesign --verify --deep --strict` cleanly after clearing destination-side Finder-added `com.apple.FinderInfo` (byte content fully faithful; the FinderInfo stamping is Finder's own copy behavior, documented in CLIPBOARD.md §6).
- ✅ **Toggle-off capped archive fallback**: with the guest File Provider disabled, a 5 MB folder pasted via the request-time archive (no FP publish in the log; real folder with contents at the destination). No archive is built at copy time in either mode.
- ✅ **Supersession/clear**: host-side `File Provider clear reconciled — superseded placeholders flushed` observed on offer supersession.

### Notes

- The xattr gap (#603) gained a second face live: copies can also *gain* `com.apple.FinderInfo` on bundle-named directories (added by Finder at the destination). Documented in CLIPBOARD.md §6 (4242439).
- Fresh-install observation (not this PR): on a brand-new guest, the agent's FP extension wasn't discoverable in System Settings until a `pluginkit -a` re-registration + agent restart; filing as a separate observed-behavior issue against the installer flow.
- Environmental note for posterity: Universal Clipboard pushes from other devices superseded offers mid-test twice; supersession behaved as designed (in-flight child fetches cancelled).

Issue #547, opened 2026-07-13
Profile the residual clipboard file-path throughput gap, then pursue SHA-256 pipelining / protobuf copy elimination if warranted


nicholas-lonsinger:
## Found during
Wrap-up of #377 (closed by the `SO_SNDBUF` + `VsockFrameDecoder` PR). This carries forward the only genuinely-open work from that issue; everything else in #377 has either shipped or been settled as a negative.

## Description
After the `SO_SNDBUF` transport unlock and the two per-chunk levers from #544, the clipboard stack still shows an unexplained gap between the inline wire+protocol ceiling (RAM both ends, ~415 MiB/s post-#544) and real file transfers (guest→host ~176 MiB/s, host→guest ~163 MiB/s). Endpoint disk I/O was measured independently and exonerated (guest virtual-disk reads and host staging writes are both 10–30× faster than any observed transfer rate), so the gap lives somewhere in the app path, not the disk.

The separating experiment is a clean **standalone** `sample` (never Xcode — it contaminates throughput) taken *during a large file transfer*, compared against a sample during an inline transfer. Named suspects from the #377 investigation: durable-write-then-ack serialization into the credit loop, File-Provider-relay overlap on the host side, and guest virtio write-side behavior (host→guest being the slower direction is consistent with that last one).

This profiling **gates** the remaining measurement-gated micro-optimizations — do not pursue them until a profile shows the term they target is actually on the critical path:

- **SHA-256 pipelining** — hash chunk N on a companion queue while N+1 writes. CLIPBOARD.md §7 forbids weakening the verify itself but explicitly sanctions pipelining. Paid on both ends, so it is a candidate on whichever side the profile implicates.
- **Protobuf encode/decode copy elimination** — the copies inside the vendored SwiftProtobuf path, i.e. the harder half of the #377 per-chunk work that the `compactIfNeeded`/slice change did not touch.

## Separate, non-code follow-up
File a Feedback with Apple for the **guest→host ~750 MiB/s transport cap**: each `VZVirtioSocketConnection` is backed by an AF_UNIX socketpair born at XNU's 8 KiB `net.local.stream.sendspace` default, and the guest→host writer's socket lives inside the `com.apple.Virtualization.VirtualMachine` helper process, so a shipping app cannot raise its send buffer. Both directions demonstrably unlock 6–8× with a larger buffer (the host→guest side is what the #377 PR ships); the guest→host side needs an OS-side fix.

[...]

PR #616, opened 2026-07-20
feat: Pipeline the clipboard receiver's staging write off the credit-ack loop

### LINKED ISSUE #615: Pipeline the clipboard stream receiver's staging write and SHA-256 out of the credit-ack loop
## Found during
Wrap-up of #547, which this issue carries forward. The profiling #547 called for is complete (results in its 2026-07-17 and 2026-07-18 comments); this is the implementation work that profiling justified.

## Description
Large clipboard file transfers plateau at ~160–200 MiB/s in both directions (multi-GiB payloads; 20 GB guest→host measured at 168.8 MiB/s) versus the ~398–415 MiB/s inline app ceiling the same stack reaches with RAM endpoints. The #547 profile located the gap: the sender spends ~92% of its busy time blocked in `OutboundTransfer.awaitCredit`, while the receiver's per-transfer serial queue alternates the staging `write(2)` (~60–73% of its busy loop) and SHA-256 update (~19–27%) before sending the credit-opening ack. Chunk receive, durable write, and hashing are fully serialized; overlapping them is the remaining sanctioned optimization.

Constraints established by prior work:
- CLIPBOARD.md §7 forbids weakening or skipping the digest verify itself but explicitly sanctions pipelining it.
- Protobuf encode/decode copy elimination was **dismissed by the same profile** (<1% in every capture) — it is settled and out of scope here.
- Rates must be measured on unsampled, standalone (non-Xcode) runs; never `sample` the `com.apple.Virtualization.VirtualMachine` helper during a timed run (throttles the transfer ~20×). See the methodology caveat on #547.

Both directions benefit: the receiver's serialization is what starves the sender's credit loop, and the split is similar host→guest and guest→host.

## Location
`ClipboardStreamReceiver.handleChunk` / `handleEnd` (per-transfer serial queue: write → hash → ack), `ClipboardFileStaging.Sink.write`; sender-side starvation visible in `ClipboardStreamSender` (`OutboundTransfer.awaitCredit`).

_Implementation/wiring intentionally omitted — design when picked up._

nicholas-lonsinger:
Closes #615

## Summary

Large clipboard file transfers plateaued at ~160–200 MiB/s in both directions against the ~398–415 MiB/s inline ceiling the same stack reaches with RAM endpoints. The #547 profile located the gap: the receiver serialized chunk receive → staging `write(2)` (~60–73% of its busy loop) → SHA-256 (~19–27%) → credit-opening ack on a single per-transfer queue, leaving the sender ~92% blocked in `OutboundTransfer.awaitCredit`.

This splits a disk-streamed transfer across **two** serial lanes so the append no longer sits between a chunk's arrival and the ack that unblocks the sender:

- **receive lane** (the existing queue) — validate, update SHA-256, refresh the stall anchor, report progress
- **write lane** (new) — append to the sink, then ack the bytes it just made durable

Chunk *n+1* is validated and hashed while chunk *n* is still being written. Both directions benefit: `ClipboardStreamReceiver` is shared KernovaKit code driven by the host `VsockClipboardService` and the guest `VsockGuestClipboardAgent`.

Hashing deliberately stays on the receive lane rather than getting a third lane: CryptoKit SHA-256 (~1.5–2 GB/s on Apple Silicon) is strictly faster than the write lane, so a separate hash lane would add a second join barrier for no throughput, and a tree hash would change the wire digest — a §7 non-starter.

## Why acks still report only durably-written bytes

Keeping `bytes_consumed = writtenBytes` is what makes the backlog self-bounding: the sender may never run more than one credit window ahead of the last acked byte, so `receivedBytes - writtenBytes` ≤ one window (1 MiB, 2 MiB at cap) per transfer. A slow volume stops the acks, the window fills, and the sender parks — real backpressure, with no new tuning constant and no un-written bytes ever counted as credit. The duplicate re-ack routes through the write lane so it carries the same durable prefix instead of racing the watermarks.

## Integrity ordering (CLIPBOARD.md §7)

Unchanged and re-documented. End verifies size + SHA-256 on the receive lane (which owns the hasher, so it already covers every accepted chunk), then commits from an async barrier queued behind the whole write backlog. Nothing is delivered and no `onTransferTimed` metric is emitted until the digest matches *and* `sink.commit()` succeeds; a digest mismatch aborts before the barrier is ever created.

## Concurrency

- Terminal state moved from a plain `finished` flag to a set-once `finishOnce()` under the transfer's own lock — completion, a failed append, and supersession now race from two lanes.
- `fail`/`finishFailed` are callable from either lane: each cleanup hops to the lane that owns it (stall timer → receive, `sink.abort()` → write), while the abort is delivered immediately so a blocked lazy pull still wakes at once even if the write lane is stuck in a slow write.
- **Bug found in review and fixed here:** `handleBegin`'s block had no terminal-state guard. A transfer becomes visible in the table *before* its Begin block is enqueued, so a concurrent `cancel(generation:)` could run teardown first and then have Begin open a sink the write lane was already aborting. That was a leaked partial plus a never-cancelled stall timer before this PR; moving `abort()` onto the write lane would have made it a genuine cross-lane data race on `sink`.
- A new `endReceived` flag drops post-End stragglers, restoring the window the single serial queue used to close implicitly.

## Test plan

- [x] `make test` — full suite green across all three targets
- [x] `make lint` clean
- [x] The four existing ack-coalescing tests are **untouched and green** — they all exercise RAM-resident inline reps, which have no write lane and keep their exact ack schedule
- [x] 7 new disk-path tests, all event-driven per docs/TESTING.md, using a new `StagingSink` seam (`GatedSink` parks an append, `FailingSink` throws on the Nth):
  - the receive lane accepts, hashes, and reports progress for every chunk while *all* writes are parked — the pipelining itself, impossible before this change
  - acks advance only as writes are released, never as bytes arrive
  - End completes only once the backlog drains and commits (asserted while the lane is provably parked inside the last write, so the negative needs no settling delay)
  - a write failing mid-backlog aborts, deletes the partial, and reports no metrics
  - a teardown with a backlog still deletes the partial
  - a volume filling mid-stream aborts from the write lane with `disk.full`
  - a duplicate chunk re-acks the *written* count while received is already ahead
- [x] Local socketpair probe (real sender/receiver, 512 MiB file rep, production sizing): 531–551 → 561–597 MiB/s, ~+8%. No regression, but a **transport-bound regime**, not the write-bound one #547 profiled — see the measurement comment below.
- [x] **Measured on the VM:** 3 GiB host→guest paste, standalone launch, agent 0.46.0, three runs with distinct payloads — 299.7 / 354.7 / 339.9 MiB/s (**mean ≈ 331**) vs the ~160–200 MiB/s plateau #547 recorded, i.e. ~**1.7×**, ahead of the projected 1.4×. Zero aborts, zero `digest.mismatch`, zero `write.short` across ~9 GiB. See the measurement comment for methodology and baseline caveats.

## Notes

- The credit window now carries a second meaning (the bound on the un-written backlog), documented in `ClipboardStreamTuning`. The receiver becomes write-lane-bound, so a deeper window buys no throughput on its own; it would only become a lever again if the write lane coalesced its backlog into one `writev(2)` per wakeup — a possible follow-up, gated on the measurement below.
- Protobuf copy elimination stays out of scope (dismissed by the #547 profile at <1% in every capture), and the sender is unchanged (`FileChunkReader.read` was 5.6% of a sender that is 92% idle).
- Guest agent bumped to 0.46.0 — receiver behavior changes inside the agent, so it needs a reinstall.

https://claude.ai/code/session_018dwiXwC4vyhcbuKRceQkfV


nicholas-lonsinger:
## Measurement status — partial, and **not** yet the production-regime number

### What I ran

A local throughput probe (untracked, not committed) driving the **real** `ClipboardStreamSender`/`ClipboardStreamReceiver` pair through the existing `StreamHarness` socketpair with production sizing (64 KiB chunks, 1 MiB window), streaming a 512 MiB **file** rep to a real staging file, reading the same `ClipboardTransferMetrics.logSummary` line the #538 telemetry emits. Three consecutive runs per branch, same machine, back to back.

| | run 1 | run 2 | run 3 |
|---|---|---|---|
| `origin/main` (27f4dfb) | 531.4 MiB/s | 535.1 MiB/s | 551.3 MiB/s |
| this branch | 561.2 MiB/s | 577.4 MiB/s | 597.2 MiB/s |

Consistent direction, non-overlapping ranges, ~**+8%**. So: **no regression, a modest gain** — but nowhere near the ~1.4× the #547 profile projects.

### Why this does not validate the projection

This harness is **not in the regime #547 profiled**. There, the receiver's busy loop was ~60–73% `write(2)` and ~19–27% SHA-256, which is what makes splitting the append off worth ~1.4×. Here both branches land at 530–600 MiB/s writing to a native APFS SSD with a warm cache, so the append is nowhere near 70% of the loop and the receiver is not the limiter — the measurement can only show the residual.

*Hypothesis (unverified — would need a profile to confirm):* the in-process sender is the limiter here. It does read + SHA-256 + send serially on one queue, and its own hash over 512 MiB is a large fraction of the ~0.9 s total. That would cap any receiver-side win regardless of size.

The conditions that make #615 matter — a **virtualized** guest disk and the real vsock transport, both far slower relative to the receive+hash lane — are exactly what a socketpair in one process cannot reproduce.

### What is still outstanding

The 3 GiB host→guest paste on a staged standalone build, read from the `.notice` telemetry lines, per #547's methodology (unsampled; never `sample` the `com.apple.Virtualization.VirtualMachine` helper). I have not run it.

> **Correction.** An earlier revision of this comment claimed this machine "has no VM library at all." **That was wrong** — retracted. The shell cannot read `~/Library/Containers/app.kernova/Data` at all (TCC: `Operation not permitted`), and the two commands I based the claim on were both silently defeated by that: a `find` with `2>/dev/null` swallowed the permission error and looked like an empty directory, and `du -sh` reported 36 KB because the only file it could traverse was the container metadata plist. Whether the VM library and the disposable `macOS TEST` VM exist is **unknown from the shell**, not disproven — and a staged build's File Provider extension was running the whole time, which points the other way.

Treat the ~1.4× as a projection from the #547 profile, not a measured result. Under CLIPBOARD.md §12 the merge decision should rest on the VM number.

### What *is* verified independently of throughput

- The pipelining is real and mechanically proven, not inferred: `receiveLaneRunsAheadOfParkedWrites` shows all four chunks validated, hashed, and progress-reported with **every** staging write parked — impossible before this change, where progress could only fire after the write returned.
- The backlog bound, the durable-ack invariant, the End barrier, and every abort/teardown path have deterministic tests.
- Full suite green; the four pre-existing ack-coalescing tests are untouched.

nicholas-lonsinger:
## Measured on the VM — the outstanding number

Ran the 3 GiB host→guest paste on a **standalone** launch (not Xcode-launched), guest agent **0.46.0**, VM `macOS TEST`, three consecutive runs with a **distinct 3 GiB payload each time** (a re-copy would hit the materialized cache and not re-transfer). Numbers are the #538 telemetry lines, read on the host via the guest log-forwarding channel:

```
transfer 131072 completed: 3221225472 bytes in 10.251 s — 299.7 MiB/s (disk, streamed 10.249 s)
transfer 196608 completed: 3221225472 bytes in  8.661 s — 354.7 MiB/s (disk, streamed  8.659 s)
transfer 262144 completed: 3221225472 bytes in  9.039 s — 339.9 MiB/s (disk, streamed  9.036 s)
```

**Mean ≈ 331 MiB/s**, against the ~160–200 MiB/s plateau #547 recorded for the same 3 GiB host→guest payload (its Phase-1 ground truth: 1.5 GiB ≈ 7.5 s ≈ 205 MiB/s).

That is roughly **1.7×** — ahead of the ~1.4× the profile projected, and it clears CLIPBOARD.md §12's "measurable win" bar. The receiver is no longer the plateau.

### Integrity held under real load

Across ~9 GiB of streamed payload: **zero** aborts, zero `digest.mismatch`, zero `write.short`. Every transfer passed the End SHA-256 verify *and* the new committed-file-size check before delivery, and all three files materialized in the guest. The new integrity check costs one `stat` per transfer and did not register in the throughput.

### Methodology, and what this comparison is not

Followed #547's constraints: unsampled runs, standalone launch (Xcode-attached runs are ~4× slower and would have invalidated this), and the `com.apple.Virtualization.VirtualMachine` helper was never sampled.

Two honest caveats on the baseline comparison:

1. **Route differs.** These runs routed through the guest's File Provider at paste (`Routed 1 file(s) through the File Provider at paste`). #547's 3 GiB host→guest figure was measured with the File Provider toggle **off**, on the synchronous path. Same receiver code either way — the transfer telemetry measures the same staging path — but it is not a same-route comparison.
2. **Different session and VM.** #547's baseline came from an earlier session; this VM was created fresh today. No old-code run was taken on *this* setup, which would have required downgrading the in-guest agent to 0.45.1.

So: a controlled same-route, same-session before/after would need the old agent installed in the guest. What is established here is that on the current build the 3 GiB host→guest paste sustains ~330 MiB/s where the documented plateau was ~160–200, with integrity intact.

Issue #668, opened 2026-07-28
Attaching a compressed disk image with a 512-aligned size fails silently


nicholas-lonsinger:
## Found during

Implementing #573, while verifying which files `VZDiskImageStorageDeviceAttachment` actually refuses.

## Description

`VZDiskImageStorageDeviceAttachment(url:readOnly:)` validates exactly one property of the file: its size is a multiple of 512 bytes. Contents are not inspected. Verified against the framework on macOS 26 (build 26A5388g) with a standalone probe:

| File | Result |
|---|---|
| 100 bytes of zeros | `VZErrorDomain` code 5, "The disk image format is not recognized." |
| 513 bytes of zeros | `VZErrorDomain` code 5 |
| 2048 bytes from `/dev/urandom` | accepted |
| 1 MiB of zeros | accepted |
| UDZO-compressed `.dmg` (11264 bytes) | accepted |
| ULFO-compressed `.dmg` | accepted |
| UDRO `.dmg` | accepted |

For the accepted compressed images the probe went further and confirmed `VZVirtualMachineConfiguration.validate()` succeeds and `VZVirtualMachine.start()` succeeds. Nothing reports an error at any stage; the guest is handed the compressed container's bytes and finds no usable filesystem.

Whether a given compressed image is caught therefore depends only on its byte size. Every `.dmg` produced by `diskutil image create from` in the probe happened to be 512-aligned; two downloaded application `.dmg` files sampled from `/Applications` were not (`mod512` of 335 and 11).

The app translates no `VZError` code 5 today — the three attach sites rethrow the framework's error unchanged (#573 is open). So an *unaligned* compressed image surfaces the framework's opaque message, and an *aligned* one produces no error at all. #573 covers the first; this issue is the second, and content inspection is the only thing that can reach it.

## Location

`ConfigurationBuilder.configureStorageDisks`, `ConfigurationBuilder.configureRemovableMedia`, `USBDeviceService.attach` — the three `VZDiskImageStorageDeviceAttachment` construction sites.

## Considerations

Catching this needs content inspection before or alongside the attach, which #573 proposes to avoid in favor of reacting to the framework's own error — that approach rests on the framework error covering the case, which this finding contradicts. Open question for whoever picks it up: whether the check reads a magic number (UDIF images carry a `koly` trailer in the last 512 bytes), and whether it warns or refuses. Design when picked up.

PR #669, opened 2026-07-28
feat: Explain and offer a fix for disk images Virtualization can't open

### LINKED ISSUE #573: feat: Translate VZ's invalid-disk-image error into an actionable message
## Summary

When a user attaches a disk image `VZDiskImageStorageDeviceAttachment` can't use (e.g. a compressed DMG), Kernova surfaces the framework's raw, opaque error instead of explaining which formats are actually supported.

## Context

- Per Apple's Virtualization framework docs, `VZDiskImageStorageDeviceAttachment` supports exactly two formats: **RAW** ("a file that's the requested size of the VM disk... a 1-to-1 mapping between the offsets in the file and the offsets in the VM disk") and **ASIF** ("size on the file system is proportional to the actual data stored"). Compressed formats are not documented as supported.
- `VZError.Code` has a dedicated `invalidDiskImage` case ("An invalid disk-image error") — the framework's own signal for this failure, distinct from other cases like `deviceInitializationFailure` or `outOfDiskSpace`.
- Reported case: a `.dmg` built in compressed (UDZO) format was rejected; repackaging the same payload as a UDRW (raw read/write) DMG with a GUID partition scheme — matching Kernova's own bundled `KernovaMacOSAgent.dmg` — resolved it, verified via `hdiutil imageinfo` (`Format: UDRW, partition-scheme: GUID`).
- The three call sites constructing `VZDiskImageStorageDeviceAttachment(url:readOnly:)` — `ConfigurationBuilder.configureStorageDisks`, `ConfigurationBuilder.configureRemovableMedia`, and `USBDeviceService.attach` — each wrap the initializer in a `do`/`catch` that logs and rethrows the framework's error unchanged. None inspect it for this case. These three sites cover both Storage Disks and Removable Media (cold-attached at VM start and hot-attached at runtime) — all three use the exact same type and initializer, so the constraint and failure mode are identical across both attachment kinds.
- `ConfigurationBuilderError` already has typed, user-facing cases for other file problems on these same paths (`storageDiskNotFound`, `storageDiskPathIsDirectory`, `storageDiskNotWritable`, and the `removableMedia*` equivalents), each rendering specific text through the existing `presentError` → `NSAlert` pipeline. There is no equivalent case for an invalid disk image today.
- Kernova's own VM-disk creation previously used `hdiutil` and was moved to bundled, pre-converted ASIF templates specifically because the App Sandbox can't usefully spawn `hdiutil`/`diskutil` and there's no public in-process API for ASIF creation. The same constraint bears on any in-app conversion of a rejected image.

## Requirements

- At each `VZDiskImageStorageDeviceAttachment` construction site, catch the `VZError.Code.invalidDiskImage` case specifically and surface a Kernova-authored message explaining that only RAW and ASIF disk images are supported — rather than pre-validating format before the attempt. Catching the framework's own error keeps the message correct if Apple adds support for another format later, with no format allowlist to maintain.
- Leave all other `VZError` cases (and non-`VZError` failures) surfaced as they are today — only `invalidDiskImage` gets the new message.
- Investigate whether in-app conversion of a rejected image is technically feasible without a subprocess or a private framework; if so, offer it as part of the failure flow — if not, provide the user instructions on how to convert their image at the command line, offering the command as a copy/paste option.

## Additional Notes

- Applies to every attach path: adding an external storage disk, adding removable media, and the runtime hot-attach path (`USBDeviceService.attach`) — not just VM start.
- Exact error copy, and whether to point users at a conversion tool (e.g. Disk Utility / `hdiutil convert`, since Kernova likely can't convert in-process), is left for whoever picks this up.

nicholas-lonsinger:
Closes #573

## Summary

When `VZDiskImageStorageDeviceAttachment` refuses a file, Kernova now says why in its own words and hands the user the command that fixes it, on every attach path — cold-attached storage disks and removable media, and the runtime USB hot-attach.

## What Virtualization actually checks

The issue's premise was that the framework rejects compressed disk images and that catching `VZError.Code.invalidDiskImage` would surface that. Verifying against the framework on macOS 26 (26A5388g) with a standalone probe showed otherwise: the initializer validates exactly one property — the file's size is a multiple of 512 bytes — and inspects no contents.

| File | Result |
|---|---|
| 100 bytes / 513 bytes of zeros | `VZErrorDomain` code 5 |
| 2 KiB from `/dev/urandom`, 1 MiB of zeros | accepted |
| UDZO / ULFO / UDRO `.dmg` | accepted |

Compressed `.dmg` files do fail in practice, but because their size is arbitrary rather than because of their format — the two application `.dmg` files sampled from `/Applications` had `mod512` of 335 and 11. Apple's own header ("Only RAW data disk images are supported") describes an intent the initializer does not enforce.

That changed the wording, not the shape of the fix: the message names the size requirement alongside the supported formats, and converting to ASIF resolves both. The residual case — an aligned compressed image that attaches with no error and is simply unreadable in the guest — needs content inspection and is filed as #668.

## In-app conversion

Not possible. Decompressing a UDIF `.dmg` needs `hdiutil`/`diskutil` or the private DiskImages framework, neither reachable from the sandbox, and ASIF creation has no public in-process API — already why disk images ship as bundled lzfse templates. So the failure flow offers the command instead: `diskutil image create from --format ASIF`, which is the non-deprecated spelling on macOS 26 (`hdiutil convert` is deprecated and cannot emit ASIF). Verified end to end that it converts a compressed source into an image the framework accepts.

## Route taken

`DiskImageFormatGuidance` is the single place that knows the VZ error code, both attach-failure messages, and the shell-quoted command. `CommandSuggestingError` lets `VMLibraryViewModel` ask any error for a copyable remedy without switching on error types.

Both `ConfigurationBuilder` sites already wrapped the failure in a typed error preserving the underlying `NSError`, so the message branches off that — no new plumbing, and the file-lock-contention retry that reads the same underlying error is untouched. `USBDeviceService.attach` was the outlier, rethrowing the framework error raw; it now wraps in `USBDeviceError.diskImageAttachFailed`. Nothing on that path inspects the raw error (all contention detection is start-path only), so the wrap is safe and it is what lets the hot-attach path carry a Kernova message at all.

The command surfaces as a selectable monospaced field with a copy button in the alert's accessory view, rather than a "Copy" button in the button row — copying doesn't dismiss the alert, so the start-failure case keeps its "Remove and Start" offer intact.

### Rejected alternatives

- **Pre-validating the format before attaching** — the issue rejected it to avoid an allowlist that would go stale if Apple widened support. Still right, though for a different reason than assumed: there is no format check to mirror.
- **A second `presentError` overload carrying the command** — one method with the parameter instead, so there is no parallel path. The ripple was four call sites.
- **Command in the message text only** — no way to copy it without retyping.

## Test plan

- [x] `make test` — 1815 cases, no failures
- [x] `make lint`
- [x] Probed the framework directly for which files it rejects, and that `diskutil image create from --format ASIF` produces one it accepts
- [x] Rendered the alert from the real view sources to confirm the accessory sizes correctly (`NSAlert` does not size accessory views)

New coverage: `DiskImageFormatGuidanceTests` (classification, both message branches, shell quoting for spaces and embedded quotes), `CopyableCommandViewTests` (pasteboard write against a private pasteboard, copy confirmation, self-sizing), plus message/`suggestedCommand` assertions on both error enums and command-propagation tests through the view model, including the buffered-before-presenter path.

## Notes

No `RATIONALE:` comments added.

### DIFF
--- a/Kernova/App/DetailContainerViewController.swift
+++ b/Kernova/App/DetailContainerViewController.swift
@@ -267,8 +267,8 @@ extension DetailContainerViewController: VMCreationWizardViewControllerDelegate 
 // MARK: - VMLibraryPresenting
 
 extension DetailContainerViewController: VMLibraryPresenting {
-    func presentError(_ message: String) {
-        alertsPresenter.presentError(message)
+    func presentError(_ message: String, copyableCommand: String?) {
+        alertsPresenter.presentError(message, copyableCommand: copyableCommand)
     }
 
     func presentStartFailedAttachment(_ failure: StartFailedAttachment, for instance: VMInstance) {
--- a/Kernova/Services/ConfigurationBuilder.swift
+++ b/Kernova/Services/ConfigurationBuilder.swift
@@ -721,7 +721,7 @@ struct ConfigurationBuilder: Sendable {
 
 // MARK: - Errors
 
-enum ConfigurationBuilderError: LocalizedError {
+enum ConfigurationBuilderError: LocalizedError, CommandSuggestingError {
     case invalidHardwareModel
     case invalidMachineIdentifier
     case missingKernelPath
@@ -773,7 +773,8 @@ enum ConfigurationBuilderError: LocalizedError {
         case .storageDiskNotWritable(let path, let label):
             "Storage disk '\(label)' is not writable: \(path). Change it to read-only or select a writable file."
         case .storageDiskAttachFailed(_, let path, let label, let underlying):
-            "Couldn't open storage disk '\(label)' at \(path). The file may have been moved or replaced, or Kernova may no longer have permission to read it. (\(underlying.localizedDescription))"
+            DiskImageFormatGuidance.attachFailureMessage(
+                subject: "storage disk '\(label)'", path: path, underlying: underlying)
         case .removableMediaNotFound(let path, let label):
             "Removable media '\(label)' not found at \(path)."
         case .removableMediaPathIsDirectory(let path, let label):
@@ -781,7 +782,8 @@ enum ConfigurationBuilderError: LocalizedError {
         case .removableMediaNotWritable(let path, let label):
             "Removable media '\(label)' is not writable: \(path). Change it to read-only or select a writable file."
         case .removableMediaAttachFailed(_, let path, let label, let underlying):
-            "Couldn't open removable media '\(label)' at \(path). The file may have been moved or replaced, or Kernova may no longer have permission to read it. (\(underlying.localizedDescription))"
+            DiskImageFormatGuidance.attachFailureMessage(
+                subject: "removable media '\(label)'", path: path, underlying: underlying)
         case .sharedDirectoryNotFound(let path):
             "Shared directory not found at \(path)."
         case .sharedDirectoryNotADirectory(let path):
@@ -799,6 +801,16 @@ enum ConfigurationBuilderError: LocalizedError {
         case .storageDiskAttachFailed(_, _, _, let underlying),
             .removableMediaAttachFailed(_, _, _, let underlying):
             underlying
+        default:
+            nil
+        }
+    }
+
+    var suggestedCommand: String? {
+        switch self {
+        case .storageDiskAttachFailed(_, let path, _, let underlying),
+            .removableMediaAttachFailed(_, let path, _, let underlying):
+            DiskImageFormatGuidance.suggestedCommand(forPath: path, underlying: underlying)
         default:
             nil
         }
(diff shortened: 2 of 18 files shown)


nicholas-lonsinger:
Backing this out for now — to be recreated later.

The message wording needs rework before this ships: `VZError.Code.invalidDiskImage` conflates a compressed container (convert it) with a truncated or damaged file (re-download it), and the version here offers the conversion command for both. Details and evidence are in [#573](https://github.com/nicholas-lonsinger/kernova/issues/573#issuecomment-5108652652); #668 tracks the related silent-acceptance case.

The full diff stays readable on this closed PR, so the work can be picked back up from here.

Issue #280, opened 2026-05-29
feat: Offer a catalog of selectable macOS restore images from Apple (not just latest)


nicholas-lonsinger:
## Summary

When creating a macOS guest, let the user pick from a **catalog of Apple's available macOS restore images** (multiple versions, optionally betas) instead of only the single newest build. Today the install flow offers exactly two choices — download latest (which resolves to `VZMacOSRestoreImage.latestSupported`, i.e. only the newest supported build) and a local file (bring-your-own IPSW). There is no way to download an *older* macOS version from within Kernova.

## Motivation

- Testing or reproducing against a specific macOS version (e.g. "install macOS 14.5") is a primary reason to run macOS VMs; pinning to "latest only" forces users to hunt down IPSW URLs by hand and use the Local File path.
- Peer apps do this: VirtualBuddy presents Apple's restore-image catalog with per-version host-support info (and betas).
- This is the **Apple-sourced** slice of the broader image-distribution idea — explicitly **not** OCI/registry pull-push, third-party Xcode-baked images, or suspended-state distribution (those are deferred).

## Scope

- Add a third install source (e.g. `.catalog`) that lists available Apple restore images and lets the user choose a version, then **reuse the existing resumable download pipeline unchanged** to fetch the chosen image.
- Source the catalog from Apple's software-update / restore-image feed (the mechanism VirtualBuddy/UTM use to enumerate versions). `VZMacOSRestoreImage.latestSupported` only ever yields one build, so a catalog requires querying Apple's `.sucatalog`/mesu feed (or equivalent) and verifying each image is Apple-signed before offering it.
- Surface the picker in the install-source step as a third radio option alongside Download Latest / Local File, annotated with version + build + host-support status; reflect the selection in the review step.

## Considerations

- Validate that a chosen guest version is ≤ host macOS and supported on the host **before** download — annotate unsupported versions in the list rather than failing late.
- Catalog fetch should be cached/refreshable and degrade gracefully offline (fall back to Download Latest / Local File).
- Keep the existing download/resume/finalize path (including its `.kernovadownload` resume bundle) as-is — only the URL-resolution front-end changes.
- **Out of scope** (track separately if wanted): custom-URL source, OCI registries, prebuilt third-party/Xcode images, Linux distro catalog.

## Found during

Competitive feature-gap analysis of VZ-based macOS-guest managers (VirtualBuddy's version catalog). Scoped per request to the Apple-image-catalog slice only.

PR #673, opened 2026-07-29
feat: Add a "Paste an IPSW URL…" install source for macOS guests

nicholas-lonsinger:
**Stacked on #672** — base is `feat/macos-restore-image-catalog-ui`, so review that one first. #280 explicitly put custom-URL out of scope, which is why this is a separate PR rather than more commits on #672.

## Summary

Adds the fourth IPSW source: install from a restore image at a URL the user supplies. It covers what the catalog structurally cannot — seeds, and any build newer than the bundled snapshot.

## The feature is the check, not the text field

Every catalog entry was verified at generation time, so the app can trust it. A pasted URL has been verified by nobody. So the sheet points the generator's own zip-central-directory read at the URL: an IPSW is a zip whose directory names `kernelcache.release.vma2` exactly when the image carries the virtual-machine hardware model, which settles it in **about 150 KB** rather than a 16 GB download that only fails at install.

That turns an unknown string into the same facts a catalog row shows. **Use** stays disabled until it passes, so nothing unchecked reaches the install pipeline.

## Any host, because the image is signed

A restore image carries `apticket.vma2macosap.im4m`, an Apple-signed IMG4 manifest that the restore validates. An image Apple did not sign cannot install, whoever served the bytes — so a host allowlist would not be the integrity boundary, only a guess at where Apple's files live, and a wrong one the day someone has a legitimate mirror. **HTTPS is the only requirement.**

Two consequences worth stating rather than discovering:

- The signature bites at *install*, after the download. The pre-download check catches the ordinary failure (not an installable image at all) for 150 KB; a well-formed image with a bad signature still surfaces at restore.
- **This does not change the catalog's host check.** The two look alike and do opposite jobs: the catalog validates data *we* generated and shipped, where an off-Apple host means the generator misbehaved and the assertion should fire. The URL source validates a string the user typed, where an unfamiliar host means nothing.

## One fact is a guess, and is treated like one

Version and build are parsed from Apple's `UniversalMac_<version>_<build>_Restore.ipsw` convention; size and VM-capability are measured. The parse yields `nil` rather than a guess for anything off-convention, and nothing that gates the install depends on it. A filename-derived version above the host **warns without blocking** — blocking on a guess would be wrong — and an unrecognized filename says so plainly instead of showing a blank.

## The layout finding from the mock, applied

The wizard sheet is fixed at 720×540, leaving the step ~371 pt. Four options plus the two chips comes to ~390 pt. Nothing breaks — the step already lives in a scroll view with a more-below cue — but scrolling a four-item radio list to reach the last option is a regression. So the image chip and the destination chip, which are two halves of one statement, collapse into **one two-line badge for every download source**, bringing it to ~346 pt. Uniform across sources rather than special-cased.

## Changes

- **`ProbedRestoreImage`** — measured facts plus the best-effort filename parse, `suggestedFilename` (per-image destination, same reason as a catalog pick), and a tri-state `isSupported(onHost:)` whose `nil` means "unknown", not "no".
- **`RestoreImageProbing` / `RestoreImageProbeService`** — `HEAD` for size, falling back to a one-byte ranged `GET`'s `Content-Range` for servers that don't answer HEAD; then the zip64-aware directory read.
- **`IPSWSource.customURL`** and `pastedImage`; **`MacOSInstallContext.Source.customURL`** with `usesPinnedURL` / `downloadsImage` computed properties replacing the case lists that were starting to spread through the coordinator and wizard.
- **`RestoreImageURLSheetContentViewController`** — 560×320 nested sheet. Editing the URL invalidates the previous verdict so a stale Use can't commit the wrong image; the in-flight probe is cancelled on dismiss and superseded rather than raced on a second Check.
- Review gains "From URL" with version and size; the initial-boot banner names the pinned image.

## Test plan

- [x] `make test` — 1865 tests, exit 0
- [x] `make lint` — clean
- [x] Probe accepts a synthetic zip naming `vma2`, refuses one that doesn't, a 404, a non-HTTPS URL (without issuing a request), and a file with no zip structure
- [x] Filename parse yields nil for every off-convention shape tried
- [x] Per-image destinations: two different URLs never resolve to the same path
- [x] Sheet: Use gated on the check, delegate callbacks, too-new warning that doesn't block, unknown-version note, edit-invalidates-verdict
- [x] `MacOSInstallContext` round-trips the new case; `usesPinnedURL` covers exactly the two pinned sources
- [ ] **Not done:** driving the wizard on screen. Computer use has been held by another session throughout. The ~346 pt figure is computed, not observed — the four options and collapsed badge fitting without scrolling is the specific thing that wants eyes.

## Notes

`ProbeStubURLProtocol` is a second stub `URLProtocol` rather than a reuse of `StubURLProtocol`. That one's handler is a global and Swift Testing runs suites in parallel, so sharing it made this suite and `IPSWServiceDownloadTests` clobber each other — both failed, then passed on retry. `.serialized` orders tests *within* a suite and does not prevent it. Caught before pushing; the isolated stub removes the flake.

https://claude.ai/code/session_01KeE8mfWnzrawSP8x3nV1JN

### DIFF
--- a/Kernova/Models/MacOSInstallContext.swift
+++ b/Kernova/Models/MacOSInstallContext.swift
@@ -9,7 +9,25 @@ struct MacOSInstallContext: Codable, Sendable, Equatable {
     enum Source: String, Codable, Sendable, Equatable {
         case downloadLatest
         case catalogVersion
+        case customURL
         case localFile
+
+        /// Whether the install downloads the image named by ``remoteURL``
+        /// rather than resolving one at Start.
+        var usesPinnedURL: Bool {
+            switch self {
+            case .catalogVersion, .customURL: true
+            case .downloadLatest, .localFile: false
+            }
+        }
+
+        /// Whether the install fetches the image over the network.
+        var downloadsImage: Bool {
+            switch self {
+            case .downloadLatest, .catalogVersion, .customURL: true
+            case .localFile: false
+            }
+        }
     }
 
     var source: Source
@@ -21,10 +39,11 @@ struct MacOSInstallContext: Codable, Sendable, Equatable {
     /// download state and enables resume across app restarts.
     var downloadDestinationPath: String?
 
-    /// The exact image to download (for `.catalogVersion`).
+    /// The exact image to download (for the sources where ``Source/usesPinnedURL``).
     ///
-    /// Pinned at wizard time from the bundled catalog, so a Start weeks later
-    /// fetches the version the user chose rather than whatever is newest.
+    /// Pinned at wizard time — from the bundled catalog, or from a URL the user
+    /// supplied and Kernova checked — so a Start weeks later fetches the image
+    /// the user chose rather than whatever is newest.
     var remoteURL: URL?
 
     /// Marketing version and build of the pinned image, for display while the
--- a/Kernova/Models/ProbedRestoreImage.swift
+++ b/Kernova/Models/ProbedRestoreImage.swift
@@ -1,0 +1,82 @@
+import Foundation
+
+/// A restore image at a user-supplied URL, after the pre-download check.
+///
+/// Only ``sizeBytes`` and the fact that the image carries the virtual-machine
+/// hardware model are measured. ``version`` and ``build`` are read out of
+/// Apple's filename convention and are absent whenever the filename does not
+/// follow it, so nothing that gates the install may depend on them.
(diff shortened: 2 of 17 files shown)


nicholas-lonsinger:
Folded into #672, which now carries both commits and targets `main`. GitHub auto-closed this as merged once `f7b9882` became reachable from its base branch — nothing was merged into `main` here.

Issue #686, opened 2026-07-31
Retarget the guest agent to macOS 12


nicholas-lonsinger:
## Summary

Lower the guest agent's deployment floor (agent app + the KernovaKit package) from macOS 26 to macOS 12.0 so every guest the app can install — the catalog reaches back to 12.0.1 — can run the agent. Today the floor is inherited from the project-level `MACOSX_DEPLOYMENT_TARGET` (no per-target override exists), the agent Info.plist interpolates it into `LSMinimumSystemVersion`, and KernovaKit declares `.macOS(.v15)` — so the built agent refuses to launch in any pre-26 guest, and `install.command` inside an old guest fails with the generic "may not be executable on this system" message.

## Verified groundwork

All of the below is verified, not estimated:

- **Transport works on macOS 12 guests.** Guest-side virtio-vsock (`AppleVirtIOSocket`) is present and completes guest-initiated connections to the host control listener end-to-end on both 12.0.1 and 12.7.6 — method and evidence in `docs/research/2026-07-31-macos12-guest-vsock.md` (merged in #685). This corrects the public "since macOS 13" claim.
- **Every API gap is macOS 13.0; nothing needs 14 or 15** (proven by `swift build` / typecheck at `arm64-apple-macos12.0`). KernovaKit: the `Duration`/`ContinuousClock`/`Task.sleep(for:)`/`Instant` family, concentrated in `ClipboardStreamTuning`, `ClipboardStreamSender`, `ClipboardStreamReceiver`, `LazyPullCoordinator`. Agent target: the same `Duration` family in `VsockGuestClient`/`VsockGuestControlAgent`, plus the no-argument `NSApplication.activate()` (macOS 14) in `AgentStatusItemController`. swift-protobuf declares no platform floor and builds at 12. Test targets (`KernovaTestSupport`, `KernovaKitTests`, `KernovaMacOSAgentTests`) use the same `Duration` family in fixtures and move with the port.
[...]

PR #702, opened 2026-08-01
feat: Retarget the guest agent to macOS 12

nicholas-lonsinger:
Implements the retarget in #686 (left open: its real-guest validation checklist is still outstanding — see Test plan).

## Summary

Lowers the deployment floor of the guest agent, its File Provider extension, KernovaKit, and KernovaMacOSAgentTests from macOS 26 to 12.0, so every macOS guest the catalog can install (back to 12.0.1) can run the agent.

Two separable bodies of work: a **clock seam** that lets the shared engine compile at the lower floor, and a set of **runtime compatibility fixes** that only a real Monterey guest surfaced.

### The clock seam

The core design is a generic clock abstraction rather than the originally-agreed single substituted path (rationale recorded in the issue's amended "Agreed approach"): the stream engine and the agent's watchdog/reconnect types are generic over a new `EngineClock` protocol with an associated `Instant` — `Swift.Clock`, one floor lower. `@available` cannot gate stored properties (a field is layout, not code), but a stored property of an associated type is legal at the 12 floor because layout resolves per instantiation. So:

- **macOS 13+ (host and modern guests)** run `ContinuousEngineClock`, storing genuine `ContinuousClock.Instant`s end-to-end — zero behavior change and no adapter code on the modern path.
- **macOS 12 guests** run `MonotonicEngineClock` on `clock_gettime_nsec_np(CLOCK_MONOTONIC)`, which counts time asleep on Darwin — matching `ContinuousClock` semantics across VM save/restore, where an uptime clock (`DispatchTime`/`mach_absolute_time`) would silently freeze the liveness watchdogs. The clockid choice lives in exactly one line.
- **Tests** get a manually advanced `TestEngineClock`, and the timing-relevant suites run parameterized over both real conformances, so the code path a 12 guest executes is exercised on every CI run.

`#available` appears only at construction sites (holders that must run on 12 store clock-free facades: `ClipboardStreamReceiving`, `VsockReconnecting`, `VsockGuestControlling`) and at genuine capability gaps: `setCodeSigningRequirement` (13+, unreachable on guests and fault-logged if ever hit), `NSApplication.activate()` vs `activate(ignoringOtherApps:)` (14), and the settings-app deep links plus "Enable in System Preferences…" copy below 13.

Rejected alternatives: call-site `#available` branching (impossible — the stored-property availability ban); a scalar type-erased seam (modern path would run epoch-conversion adapter code instead of native instants); forking the engine into parallel `@available` types (duplicates ~800 lines of the subtlest concurrency code). When the 12 floor is eventually dropped, `EngineClock` collapses onto `Swift.Clock` by deleting the legacy conformance and the construction branches.

`ClipboardStreamSender` and `LazyPullCoordinator` turned out to need no clock at all — their `Duration`s were semaphore/condition deadlines, now `TimeInterval` (identical wait mechanics). Boundary currency (tuning constants, transfer metrics, facade surfaces) is `TimeInterval` throughout.

### Runtime compatibility

Lowering the deployment target surfaces none of what follows: it was all found by running the agent on a 12.7.6 guest, and each finding carries a dated research note with its method. The `getService` case is the sharp one — its availability is declared on the Objective-C *category*, which the Swift importer does not surface, so `swiftc -typecheck -target arm64-apple-macos12.0` accepts the call with zero diagnostics and Monterey aborts on the selector at runtime. Every other guest-reachable FileProvider symbol was audited against the same header.

The `getService` abort presented as the agent's status icon disappearing: a crash loop on launchd's 10-second throttle, starting the instant clipboard sharing was enabled, because that policy is what arms the servicing connector. Fixing it exposed the parked connect underneath. Through both, the host's liveness watchdog logged the channel silent every 5 seconds without ever tearing it down or clearing `isConnected` — pre-existing on `main`, so filed as #703 rather than widened into this PR.

## Changes

**Floor and the clock seam**

- New: `EngineClock.swift`, `ClipboardStreamReceiving.swift` (KernovaKit); `TestEngineClock.swift`, `EngineClockKind.swift` (KernovaTestSupport)
- `ClipboardStreamReceiver` generic over the clock (all four stored timestamps are `Clock.Instant`); sender/coordinator/tuning move to `TimeInterval`
- `VsockGuestClient` and `VsockGuestControlAgent` generic over the clock; `AgentAppDelegate`, `VsockHostConnection`, `VsockGuestClipboardAgent` hold facades; `classifySocketErrno` hoisted to file scope (a generic type can't hold static stored properties)
- The reported OS version parses without Swift Regex (13+)

**macOS 12 runtime compatibility**

- `openVsockToHost` connects blocking below macOS 13: Monterey's virtio-vsock driver never completes a non-blocking `connect(2)` (`docs/research/2026-08-02-macos12-vsock-nonblocking-connect.md`). The macOS 13+ path keeps the non-blocking-plus-poll idiom unchanged
- That blocking connect is bounded at the same `connectTimeoutSeconds` ceiling by running it on a throwaway thread the loop abandons: Darwin bounds `connect(2)` with no socket option, and a vsock connect does park indefinitely when the host holds a listener that never accepts (`docs/research/2026-08-02-macos12-vsock-blocking-connect-parks.md`). The abandoned worker keeps its socket and closes it when the kernel returns, so nothing is closed under an in-flight syscall, and `BlockingConnectGate` admits one connect per label so a stalled host cannot accumulate a thread per retry
- The File Provider servicing connector addresses its service by the domain root's user-visible URL below macOS 13: `getService(named:for:)` is a macOS 13 selector that Swift compiles ungated at the 12 floor, so a Monterey guest aborted on it the moment clipboard sharing was enabled (`docs/research/2026-08-02-macos12-fileprovider-getservice.md`)
- The "Enable in System Preferences…" command reaches the Extensions pane by opening the pane bundle's own file URL below 13: `com.apple.preferences.extensions` is the pane's real bundle identifier and System Preferences still does not route it, so the command used to leave the user at the main grid (`docs/research/2026-08-03-macos12-extensions-pane-deep-link.md`). The OS-era split moved into `enablementDeepLinks(systemSettings:)` so both candidate lists are testable from either OS

**Build**

- Per-target `MACOSX_DEPLOYMENT_TARGET = 12.0` (agent, agent FP extension, agent tests — project level stays 26); KernovaKit `platforms: .macOS(.v12)`; agent `MARKETING_VERSION` 0.52.0 → 0.53.0 in all four places (rebased over #696's own 0.52.0 bump) (forces guest reinstall per BUILD.md)
- `LD_RUNPATH_SEARCH_PATHS` on the agent app target, which had none and so inherited a list without `@executable_path/../Frameworks` — when Xcode links the packages dynamically the agent embeds frameworks it cannot resolve, and `install.command` aborts at its `--version` probe. The agent's own File Provider extension already declared both paths

**Tests and docs**

- Clock parameterization via `EngineClockKind`; `KernovaMacOSAgentTests` verified free of any macOS 13+ API; `AsyncWaits`/`EphemeralDefaults` made 12-safe
- `ClipboardFileProviderSettingsTests` pins that the deep-link candidate able to fail honestly stays ahead of the `x-apple.systempreferences:` one that reports success regardless
- README and ARCHITECTURE state host-vs-guest floors separately; ARCHITECTURE notes the clock seam and facades

## Test plan

- [x] `make test` green: host, agent, and package targets (2076 cases; the package built at the 12.0 floor)
- [x] KernovaKit + KernovaTestSupport compile at an `arm64-apple-macos27.0` target under `-warnings-as-errors` with zero diagnostics — the modern-end cleanliness gate from the issue
- [x] `make lint` green
- [x] 12.7.6 real guest: `install.command` succeeds and the control handshake completes and holds — host logs `Guest agent connected (service=1, agent=0.53.0, caps=control.v1,control.heartbeat.v1,clipboard.stream.v1,clipboard.dirtree.v1)` with no reconnect churn after
- [x] 12.7.6 real guest, clipboard sharing enabled: agent survives (no crash reports, PID stable), guest and host File Provider servicing connections both establish, host→guest passthrough verified end to end — a marker string put on the host pasteboard came back from `pbpaste` in the guest — and the control channel held 26 consecutive heartbeats across a mid-session clipboard off→on toggle, the sequence that first exposed all of this
- [x] 12.7.6 real guest, File Provider enablement: the status-item "Enable in System Preferences…" command lands on Extensions → Added Extensions with the agent's checkbox in view, and checking it there moves the guest's availability from `needsEnabling` to `ready`
- [ ] Remaining real-guest validation (issue #686, requirements 3–4): 12.0.1 and a modern guest, guest→host clipboard direction, File Provider file transfer end to end on 12 (app-group container sharing verified by data flow, stabilization behavior), and the save/restore watchdog pass — the only step that executes `MonotonicEngineClock` under the time-asleep condition the clockid choice exists for
- [ ] SANDBOX.md criterion-C qualification per what that testing shows (issue #686, requirement 5)

https://claude.ai/code/session_01QADQnA4BCWirzxtTWNdBTw

### DIFF
--- a/Kernova.xcodeproj/project.pbxproj
+++ b/Kernova.xcodeproj/project.pbxproj
@@ -1012,7 +1012,8 @@
 				INFOPLIST_FILE = KernovaMacOSAgent/Info.plist;
 				INFOPLIST_PREFIX_HEADER = "$(DERIVED_FILE_DIR)/AgentBuildNumber.h";
 				INFOPLIST_PREPROCESS = YES;
-				MARKETING_VERSION = 0.52.0;
+				MACOSX_DEPLOYMENT_TARGET = 12.0;
+				MARKETING_VERSION = 0.53.0;
 				EXECUTABLE_NAME = KernovaMacOSAgent;
 				PRODUCT_BUNDLE_IDENTIFIER = app.kernova.macosagent;
 				PRODUCT_MODULE_NAME = KernovaMacOSAgent;
@@ -1038,7 +1039,8 @@
 				INFOPLIST_FILE = KernovaMacOSAgent/Info.plist;
 				INFOPLIST_PREFIX_HEADER = "$(DERIVED_FILE_DIR)/AgentBuildNumber.h";
 				INFOPLIST_PREPROCESS = YES;
-				MARKETING_VERSION = 0.52.0;
+				MACOSX_DEPLOYMENT_TARGET = 12.0;
+				MARKETING_VERSION = 0.53.0;
 				EXECUTABLE_NAME = KernovaMacOSAgent;
 				PRODUCT_BUNDLE_IDENTIFIER = app.kernova.macosagent;
 				PRODUCT_MODULE_NAME = KernovaMacOSAgent;
@@ -1054,6 +1056,7 @@
 				CURRENT_PROJECT_VERSION = 1;
 				DEAD_CODE_STRIPPING = YES;
 				GENERATE_INFOPLIST_FILE = YES;
+				MACOSX_DEPLOYMENT_TARGET = 12.0;
 				MARKETING_VERSION = 0.9.1;
 				PRODUCT_BUNDLE_IDENTIFIER = app.kernova.macosagent.tests;
 				PRODUCT_NAME = "$(TARGET_NAME)";
@@ -1069,6 +1072,7 @@
 				CURRENT_PROJECT_VERSION = 1;
 				DEAD_CODE_STRIPPING = YES;
 				GENERATE_INFOPLIST_FILE = YES;
+				MACOSX_DEPLOYMENT_TARGET = 12.0;
 				MARKETING_VERSION = 0.9.1;
 				PRODUCT_BUNDLE_IDENTIFIER = app.kernova.macosagent.tests;
 				PRODUCT_NAME = "$(TARGET_NAME)";
@@ -1096,7 +1100,8 @@
 					"@executable_path/../Frameworks",
 					"@loader_path/../Frameworks",
 				);
-				MARKETING_VERSION = 0.52.0;
+				MACOSX_DEPLOYMENT_TARGET = 12.0;
+				MARKETING_VERSION = 0.53.0;
 				PRODUCT_BUNDLE_IDENTIFIER = app.kernova.macosagent.fileprovider;
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SKIP_INSTALL = YES;
@@ -1127,7 +1132,8 @@
 					"@executable_path/../Frameworks",
 					"@loader_path/../Frameworks",
 				);
-				MARKETING_VERSION = 0.52.0;
+				MACOSX_DEPLOYMENT_TARGET = 12.0;
+				MARKETING_VERSION = 0.53.0;
 				PRODUCT_BUNDLE_IDENTIFIER = app.kernova.macosagent.fileprovider;
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SKIP_INSTALL = YES;
--- a/Kernova.xcodeproj/project.pbxproj
+++ b/Kernova.xcodeproj/project.pbxproj
@@ -1012,6 +1012,11 @@
 				INFOPLIST_FILE = KernovaMacOSAgent/Info.plist;
 				INFOPLIST_PREFIX_HEADER = "$(DERIVED_FILE_DIR)/AgentBuildNumber.h";
 				INFOPLIST_PREPROCESS = YES;
+				LD_RUNPATH_SEARCH_PATHS = (
+					"$(inherited)",
+					"@executable_path/../Frameworks",
+					"@loader_path/../Frameworks",
+				);
 				MACOSX_DEPLOYMENT_TARGET = 12.0;
 				MARKETING_VERSION = 0.53.0;
 				EXECUTABLE_NAME = KernovaMacOSAgent;
@@ -1039,6 +1044,11 @@
 				INFOPLIST_FILE = KernovaMacOSAgent/Info.plist;
 				INFOPLIST_PREFIX_HEADER = "$(DERIVED_FILE_DIR)/AgentBuildNumber.h";
 				INFOPLIST_PREPROCESS = YES;
+				LD_RUNPATH_SEARCH_PATHS = (
+					"$(inherited)",
+					"@executable_path/../Frameworks",
+					"@loader_path/../Frameworks",
+				);
 				MACOSX_DEPLOYMENT_TARGET = 12.0;
 				MARKETING_VERSION = 0.53.0;
 				EXECUTABLE_NAME = KernovaMacOSAgent;
--- a/Kernova/Services/VsockClipboardService.swift
+++ b/Kernova/Services/VsockClipboardService.swift
@@ -55,7 +55,7 @@ final class VsockClipboardService: ClipboardServicing {
 
     /// Backstop for a lazy pull the peer never answers while the channel stays
     /// open.
-    private let lazyPullTimeout: Duration
+    private let lazyPullTimeout: TimeInterval
 
     /// Off-main authority for this VM's clipboard progress, aggregating every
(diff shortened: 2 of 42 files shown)


nicholas-lonsinger:
Real-guest validation on the 12.7.6 library VM found the agent could never connect — and the cause predates this PR's diff: Monterey's virtio-vsock driver never completes a `connect(2)` issued on an `O_NONBLOCK` socket. `poll` raises `POLLHUP` with a clean `SO_ERROR` instantly, `getpeername` stays `ENOTCONN` through the whole deadline, and the host's accept fires only when the guest closes the fd (the queued request flushes at teardown), so a host-side accept looked deceptively like progress. #685's probe connected *blocking*, which works fine — the transport itself is healthy.

Fixed in e24734d: the non-blocking-plus-poll idiom is gated to macOS 13+ (unchanged there) and Monterey connects blocking — safe on a local-only transport where the hypervisor answers accept or refusal promptly. Full observations and method: `docs/research/2026-08-02-macos12-vsock-nonblocking-connect.md`.

After the fix, on 12.7.6: `Guest agent connected (service=1, agent=0.53.0, caps=control.v1,control.heartbeat.v1,clipboard.stream.v1,clipboard.dirtree.v1)`, policy update delivered, channel holds with zero reconnect churn.

nicholas-lonsinger:
Real-guest validation on the 12.7.6 guest turned up two defects behind "the agent status icon disappeared after enabling clipboard sharing". Both are fixed here; a third, pre-existing one is filed as #703.

**1. The agent crash-looped the moment clipboard sharing was enabled.** `FileProviderServicingConnector`'s connect operation calls `NSFileProviderManager.getService(named:for:)`, whose category is declared `API_AVAILABLE(macos(13.0))`. The Swift importer does not surface that category-level availability — `swiftc -typecheck -target arm64-apple-macos12.0` accepts the ungated call with zero diagnostics — so nothing in the build caught it, and Monterey raised an unrecognized selector and `abort()`ed on the connector queue. Because the servicing connector is armed from the clipboard policy, the guest ran fine for 25 minutes under `clipboard=false` and died within seconds of the toggle, relaunching on launchd's 10 s `ThrottleInterval` — which is what made the status icon vanish and reappear.

Below 13 the connector now resolves the domain root's user-visible URL and uses `FileManager.getFileProviderServicesForItem(at:)` (macOS 10.13+). I audited every other FileProvider symbol the agent-side code reaches against the ObjC availability macros: this selector was the only 13.0 declaration among them.

**2. The blocking `connect(2)` from the previous commit could park forever.** With the crash fixed, the channel dropped once during the File Provider standup and the agent then never reconnected — `sample` showed all 1491 samples of the reconnect thread in `__connect`, parked for 12+ minutes and cleared only by SIGKILL. Darwin bounds `connect(2)` with no socket option, so the blocking call I added had no ceiling at all where the 13+ path has 3 s.

It now runs on a throwaway thread the loop abandons at `connectTimeoutSeconds`. The abandoned worker keeps its socket and closes it when the kernel finally returns, so no descriptor is closed while a syscall still holds it, and one connect per client label is admitted so a permanently stalled host cannot accumulate a thread per retry.

**3. Filed as #703 (pre-existing on `main`, not touched here).** `VsockControlService.checkLiveness()` closes a silent channel but never cancels its own liveness task, and `consume` never wires channel death back to `stop()` — so the host logged "silent for N sec — closing" every 5 s with N growing past 747, and `isConnected` stayed `true` for the rest of the session. The guest already has the wiring this is missing.

**Verification on the 12.7.6 guest, with the fixes installed:** no new crash reports and a stable PID; the control channel, log forwarding, clipboard service, and File Provider servicing connections on *both* sides all stand up; the channel held 26 consecutive heartbeats across a mid-session clipboard off→on toggle — the exact sequence that first broke it; and a marker string placed on the host pasteboard came back verbatim from `pbpaste` inside the guest.

nicholas-lonsinger:
Superseded by #747, which ports this work onto the post-File-Provider codebase: the clock seam and the vsock Monterey fixes carry over intact, the File Provider compatibility fixes are moot now that the subsystem is removed, and the boundary-currency conversions were re-derived against the #714 promise-engine re-architecture. The `feat/agent-macos12-floor` branch stays on origin until #747 merges.

Issue #706, opened 2026-08-03
Agent status sits at .connecting after a mid-session control-channel death, with no escalation to .expectedMissing


nicholas-lonsinger:
## Summary

`VsockControlService` now tears itself down when the control channel dies (#703), so `agentStatus` drops to `.waiting` and `AgentStatus.synthesize` renders `.connecting(expected:)` for any VM with a `lastSeenAgentVersion` in a live session. That is correct while the guest agent is reconnecting, but two things about the `.connecting` presentation only hold for the post-boot case it was written for.

## Observed

`AgentStatusPopoverContentViewController.bodyText` for `.connecting` reads:

> Waiting for guest agent \<version\> on \<vm\> to reconnect **after boot**. If it doesn't connect within a couple of minutes, you'll see a 'didn't reconnect' indicator with reinstall steps.

After a mid-session death (guest agent quit, or the liveness watchdog terminated a silent channel):

- "after boot" does not describe what happened — the agent connected earlier in this session and then went away.
- The promised escalation does not arrive. `VMInstance.startAgentPostStartWatchdog` is one-shot per session — it guards on `agentPostStartTask == nil` and is cancelled by the first `Hello` through `recordObservedAgentInfo` — so nothing re-arms it, and `.expectedMissing` is unreachable for the rest of the session. The badge stays at "Connecting to guest agent" indefinitely when the agent cannot come back.

A live-paused VM reaches the same state by design: the guest is frozen, the host-side liveness watchdog keeps ticking, and the channel is terminated for silence. Recovery on resume is automatic.

## Scope

The naive fix — re-arming the post-start watchdog whenever the service settles — is wrong on its own: a live-paused VM would settle after the terminate window and then fire `.expectedMissing` on a VM that is merely paused, and the watchdog's config mutation clears `agentInstallNudgeDismissed` (a user preference that nothing restores) alongside `lastSeenGuestOSVersion`. Suspending the liveness watchdog while the VM is live-paused looks like a prerequisite. Design when picked up.

## Location

- `AgentStatus.synthesize`, `AgentStatusPopoverContentViewController.bodyText` / `title`, `SidebarAgentStatusButtonView`
- `VMInstance.startAgentPostStartWatchdog`, `VMInstance.recordObservedAgentInfo`
- `VsockControlService.stop()`, `VsockControlService.checkLiveness()`

PR #707, opened 2026-08-03
fix: Stop the control-channel watchdog once it closes a silent channel

### LINKED ISSUE #703: Host control-channel liveness watchdog never stops after it closes a silent channel
## Summary

`VsockControlService.checkLiveness()` closes a control channel that has gone silent past `terminateAfter`, but nothing tears the service down afterwards. The watchdog keeps re-firing every `livenessTickInterval` (5 s) against the same frozen `lastInboundFrame`, so the host logs a growing "silent for N sec — closing" line forever, and `isConnected` stays `true` for the rest of the VM session.

Observed on a macOS 12.7.6 guest whose agent could not reconnect: the host logged the line at 46, 51, 57 … 747 seconds and beyond, once every 5 s, with no further state change.

## Why it persists

Three gaps combine:

- `checkLiveness()` calls `channel.close()` but does not cancel `livenessTask` or `outboundHeartbeatTask`, does not clear `lastInboundFrame`, and sets no already-closed latch. `VsockChannel.teardown` early-returns on its `guard !closed`, so every call after the first is a no-op while the timer keeps ticking.
- `VsockControlService.consume(channel:label:dispatch:)` has no completion callback. When `incoming` finishes it logs and the task simply ends — nothing calls `stop()`.
- `stop()` is the only path that cancels those tasks and resets `isConnected`, and it is reached only from `VMInstance.startVsockServices()`'s `onConnect` (that is, on the *next* accept) or from `stopVsockServices()`. The design assumes the guest reconnects; when it cannot, the service never settles.

## Impact

While it is stuck: `isConnected` misreports a live control channel, `admitsFeatureChannel` keeps admitting log and clipboard channels against a dead one, and every heartbeat throws `VsockChannelError.closed` (`error 1`) every 5 s. `agentStatus` does still report `.unresponsive`, because `isUnresponsive` latches during the 15–30 s window.

## Where

- `Kernova/Services/VsockControlService.swift` — `checkLiveness()`, `consume(channel:label:dispatch:)`, `stop()`, the liveness task
- `Kernova/Models/VMInstance.swift` — `startVsockServices()` / `stopVsockServices()`, the only `stop()` callers
- `KernovaKit/Sources/KernovaKit/VsockChannel.swift` — `teardown`'s closed guard

## Notes

Pre-existing on `main` — `git diff origin/main...HEAD` over these files is empty on the branch that surfaced it (#702, the macOS 12 agent retarget). The guest side already has the wiring this is missing: `VsockGuestControlAgent.serve` cancels its heartbeat and liveness tasks in a `defer` when the `incoming` loop unwinds.

Found while validating #686 on a real Monterey guest; the guest-side cause of the missed reconnect is fixed in #702, which is what stops this from being reachable that particular way.

nicholas-lonsinger:
## Summary

`VsockControlService.checkLiveness()` closed a control channel that had gone silent past `terminateAfter` and then left everything else running. `VsockChannel.teardown` early-returns on its closed guard, so every later close was a no-op while the liveness task kept ticking against a frozen `lastInboundFrame` — the host logged "silent for N sec — closing" every 5 s forever, `isConnected` stayed `true` (so `admitsFeatureChannel` kept letting log and clipboard channels onto a dead channel), and each heartbeat threw `.closed`.

The service now settles itself from **either** death path:

- **Liveness timeout** — `checkLiveness()`'s terminate branch calls `stop()` instead of `channel.close()`.
- **Consume loop ending** — the consume task calls `stop()` once `incoming` finishes, which is what was missing on a plain EOF.

A `hasStopped` latch makes the teardown happen exactly once no matter which of the three callers (watchdog, consume task, owner) arrives first. Reconnects are unaffected: the listener host is untouched, and `VMInstance.startVsockServices()`'s accept closure already builds a fresh service per connection, so the settled instance is simply replaced.

Calling `stop()` from inside the tasks it cancels is safe, and that is the load-bearing detail: `Task.cancel()` only sets the cancellation flag, and `stop()` has no suspension point after the self-cancel, so it runs to completion and the caller's loop exits at its own next `Task.isCancelled` check. That contract is now on `stop()`'s doc comment.

Closes #703

## Changes

- `Kernova/Services/VsockControlService.swift` — `hasStopped` latch on `stop()`; `checkLiveness()` and the consume task both reach it. `start()` refuses to restart a stopped service and `handle(frame:)` drops frames buffered past the teardown, so the settled state stays terminal. Adds a `#if DEBUG` `lifecycleTasksForTesting` seam.
- `KernovaTests/VsockControlServiceTests.swift` — `terminateClosesChannel` becomes `terminateSettlesTheService`; new `consumeEndSettlesTheService` and `settledServiceStaysDown`.
- `KernovaTests/VMInstanceVsockAdmissionTests.swift` — new `settleThenReconnectRestoresAdmission`.

Tests await the production tasks' completion (docs/TESTING.md's third seam) rather than asserting on `isConnected` alone — a fix that reset the flag but left the watchdog looping would pass an `isConnected`-only test.

## Deviations from what the issue proposed

- **The latch went on the service, not the channel.** The issue points at `VsockChannel.teardown`'s closed guard; that guard is correct as-is. The missing latch is `hasStopped` on `VsockControlService`, which is what makes the teardown run exactly once across three possible callers.
- **No mirror of the guest's `defer`-based structure.** The issue cites `VsockGuestControlAgent.serve`'s `defer { heartbeatTask.cancel(); livenessTask.cancel() }`. The host's tasks are unstructured and owned by the service rather than by a `serve` frame, so the equivalent is `stop()` reached from both death paths — a `defer` inside the consume task would cover EOF only, missing the liveness timeout that the issue actually observed.
- **Feature channels are deliberately excluded.** `consume` is a private static per service, not shared. `VsockGuestLogService` holds no observable state, and `VsockClipboardService`'s consume task already runs a targeted completion (`receiver.cancelAll()`, `lazyCoordinator.failAll()`); promoting that to a full `stop()` would ref-count down the shared File Provider domain while a paste may still be materializing. Neither runs a periodic task, so neither has this bug.
- **No `VMInstance` change.** `VsockControlService` is `@Observable` and the status UI already tracks `agentStatus`, so the teardown propagates with no wiring. An `onSettled` callback clearing `vmInstance.vsockControlService` would be actively wrong: a settle can land *after* the accept path installed a replacement, and would null out the fresh service.

## Behavior change worth flagging

`agentStatus` goes from a permanent `.unresponsive(version:)` to `.connecting(expected:)` at teardown. Preserving `.unresponsive` was implemented first and then rejected: the watchdog cannot distinguish a hung agent from a **live-paused VM** — a >30 s pause goes silent the same way — so latching a warning-colored "Guest agent unresponsive" badge past teardown asserts a cause the service does not know. A paused VM now gets the gray connecting state and recovers to `.current` on resume.

The cost is that `.connecting`'s popover copy promises a "didn't reconnect" escalation that the one-shot post-start watchdog will not deliver mid-session. Filed as #706 rather than fixed here: re-arming that watchdog naively would fire a false `.expectedMissing` on any live-paused VM and silently clear the user's `agentInstallNudgeDismissed` preference.

## Test plan

- [x] `make build`
- [x] `make test` — all three targets green
- [x] `make lint`

### DIFF
--- a/Kernova/Services/VsockControlService.swift
+++ b/Kernova/Services/VsockControlService.swift
@@ -25,17 +25,20 @@ struct ObservedAgentInfo: Equatable, Sendable {
 /// `VZVirtioSocketDevice`, independent of any feature toggle, and carries the
 /// bidirectional `Hello` handshake plus a `Heartbeat` stream where extended
 /// peer silence means the peer is hung. One instance manages one channel for
-/// one accepted connection; `stop()` is idempotent.
+/// one accepted connection and stops itself once that channel dies, whether the
+/// peer closed it or the watchdog did; `stop()` is idempotent and terminal, so
+/// a reconnect is served by a fresh instance.
 @MainActor
 @Observable
 final class VsockControlService {
     // MARK: - Observable state
 
-    /// `true` once the guest agent has sent its `Hello`; reset on `stop()`.
+    /// `true` once the guest agent has sent its `Hello`; reset when the
+    /// connection settles.
     private(set) var isConnected: Bool = false
 
     /// The guest-reported `Hello.agent_info.agent_version`, `nil` until that
-    /// `Hello` arrives and again after `stop()`.
+    /// `Hello` arrives and again once the connection settles.
     private(set) var agentVersion: String?
 
     /// `true` when the inbound liveness watchdog has fired but the channel has not yet been torn down.
@@ -76,6 +79,21 @@ final class VsockControlService {
     private var consumeTask: Task<Void, Never>?
     private var outboundHeartbeatTask: Task<Void, Never>?
     private var livenessTask: Task<Void, Never>?
+
+    /// Guards the teardown against re-entry: the liveness watchdog, the consume
+    /// task and the owner can each reach `stop()`, and whichever arrives first
+    /// is the one that settles.
+    private var hasStopped = false
+
+    #if DEBUG
+    /// The tasks `start()` spun up, so a test can await their completion instead
+    /// of polling for the watchdog to fall silent.
+    ///
+    /// Read it before the teardown that clears them.
+    var lifecycleTasksForTesting: [Task<Void, Never>] {
+        [consumeTask, outboundHeartbeatTask, livenessTask].compactMap { $0 }
+    }
+    #endif
 
     /// Instant of the most recent inbound frame of any kind — `Hello` and
     /// `Heartbeat` both count as liveness signals.
@@ -145,7 +163,7 @@ final class VsockControlService {
     // MARK: - Lifecycle
 
     func start() {
-        guard consumeTask == nil else { return }
+        guard consumeTask == nil, !hasStopped else { return }
 
         sendHello()
 
@@ -155,6 +173,9 @@ final class VsockControlService {
             await Self.consume(channel: channel, label: label) { @MainActor frame in
                 self?.handle(frame: frame)
             }
+            // The channel is gone once `consume` returns; settle so the
+            // heartbeat and liveness tasks stop running against a dead socket.
+            self?.stop()
         }
 
         let heartbeatInterval = self.heartbeatInterval
@@ -186,7 +207,15 @@ final class VsockControlService {
         Self.logger.info("Vsock control service started for '\(self.label, privacy: .public)'")
     }
 
+    /// Tears the service down and resets the handshake state.
+    ///
+    /// Safe to call from inside the tasks it cancels: `Task.cancel()` only sets
+    /// the cancellation flag, so a caller running inside the liveness or consume
+    /// task runs this method to completion and then unwinds at its own next
+    /// cancellation check.
     func stop() {
+        guard !hasStopped else { return }
+        hasStopped = true
         consumeTask?.cancel()
         consumeTask = nil
         outboundHeartbeatTask?.cancel()
@@ -294,9 +323,7 @@ final class VsockControlService {
             Self.logger.warning(
                 "Control channel for '\(self.label, privacy: .public)' silent for \(elapsed.formatted(.units(allowed: [.seconds])), privacy: .public) — closing"
             )
-            // Tearing the channel down lets the consume task return and the
-            // listener accept a fresh connection.
-            channel.close()
+            stop()
         } else if elapsed > unresponsiveAfter {
             if !isUnresponsive {
                 Self.logger.warning(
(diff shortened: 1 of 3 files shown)


nicholas-lonsinger:
### Code review

Reviewed the PR diff (medium effort, precision-biased per [CLAUDE.md](https://github.com/nicholas-lonsinger/kernova/blob/79810cc4a9c534220b5934d98f84585d871d9eb1/CLAUDE.md) "Matching review effort to the diff").

**Verified correct**

- *Self-cancellation.* `checkLiveness()` and the consume task both call `stop()` from inside tasks `stop()` cancels. `Task.cancel()` only raises the flag, and `stop()` has no suspension point after the self-cancel, so it completes and the caller's `while !Task.isCancelled` exits. Nulling the handle mid-execution is safe — an unstructured `Task` is retained by the runtime, not the handle.
- *Accept-vs-settle ordering.* A settle can land after the accept path installed a replacement service. Because the settle touches only its own instance and never `VMInstance.vsockControlService`, both orderings are safe. The two `stop()` call sites (`startVsockServices()`'s accept closure, `stopVsockServices()`) each replace or nil the service afterwards, so the terminal latch cannot strand a reusable instance.
- *Teardown ordering is load-bearing.* `stop()` runs `channel.close()` before `isConnected = false`, which is what makes `terminateSettlesTheService`'s trailing throws-assertion ordered rather than racy. The test comment records that coupling.
- *No orphaned observable state.* `isUnresponsive` has no consumer outside `agentStatus`, so resetting it uniformly cannot desync another reader.

**Findings**

1. Fixed in 79810cc — `settledServiceStaysDown` started a service with no matching `defer { service.stop() }`, the only test in the suite doing so; an early throw would have leaked its heartbeat and liveness tasks.

**Dismissed** (below the severity bar in [docs/REVIEW.md](https://github.com/nicholas-lonsinger/kernova/blob/79810cc4a9c534220b5934d98f84585d871d9eb1/docs/REVIEW.md))

- The terminate log line still reads "— closing" although the branch now stops the whole service. Cosmetic, and the message still describes what the user-visible effect is.
- `#expect(instance.vsockControlService !== first)` is close to tautological, since the test assigns `second` itself. Harmless, and the preceding admission wait is what actually proves the reconnect handshook.

**Deferred**

- #706 — `agentStatus` now lands on `.connecting(expected:)` after a mid-session channel death, whose popover copy promises a "didn't reconnect" escalation the one-shot post-start watchdog will not deliver. Not fixed here: re-arming that watchdog naively would fire a false `.expectedMissing` on any live-paused VM.

No issues found in CLAUDE.md / AGENTS.md adherence: no new `RATIONALE:` (one was removed, disclosed in the commit's `## Notes`), no persisted-format change, no `print`/force-unwrap, and the `#if DEBUG` test seam follows [docs/TESTING.md](https://github.com/nicholas-lonsinger/kernova/blob/79810cc4a9c534220b5934d98f84585d871d9eb1/docs/TESTING.md) "Test-only seams" option 2.

PR #738, opened 2026-08-06
fix: Roll a prefixed version-pinned catalog directory forward on regen

### LINKED ISSUE #728: Teach regen-linux-image-catalog's pinned-directory bump the kali-YYYY.N shape
## Found during
Applying review fixes on PR #725 — repinning the Kali catalog entry to its versioned directory.

## Description
The Kali entry now points at a version-pinned directory (kali-2026.2/) so the picker's version label matches what downloads. The regen tool's pinned-directory refresh only recognizes segments that split on "." into exactly three numeric parts (the Debian 12.15.0 archive shape), so a kali-YYYY.N segment is never bumped. When Kali retires the directory (it keeps roughly two quarters), a regen run hard-fails on the unreachable manifest and the entry needs a hand edit. Verified in code: pinnedVersionDirectory returns nil for the Kali URL.

## Location
`pinnedVersionDirectory` / `newerPinnedDirectory` in Tools/regen-linux-image-catalog.swift; the Kali entry in Kernova/Resources/LinuxImageCatalog.json.

## Suggested fix
Recognize the kali-YYYY.N segment shape (same-host, same-prefix substitution rule as the Debian case) so a routine regen rolls the pin forward.

nicholas-lonsinger:
## Summary

The regen tool's pinned-directory refresh only recognized a path segment that split on `.` into exactly three numeric parts — the Debian `12.15.0/` shape — so the Kali entry's `kali-2026.2/` never parsed as a pin. Kali keeps roughly two quarters of directories, so a run made after it retires one hard-failed on the unreachable manifest and the entry needed a hand edit.

Recognizing the shape was necessary but not sufficient: **bumping `directoryURL` alone would have traded one hard failure for another.** The Kali entry's `isoPattern` (`kali-linux-2026.2*-installer-arm64.iso`) and `version` label both restate the pinned version, and a bumped directory serves no filename matching the old one — the run would have failed at glob-matching instead of at the manifest fetch. So every field restating the pin now follows it, each reported as a note for the reviewed diff.

That raised the question of how far a bump may travel, since the two pinned entries want different things. **The entry's own version label states it:** Debian's `"12"` names the series the pin sits inside — the archive lists 13.x.y right beside 12.x.y, so the major has to hold — while Kali's `"2026.2"` *is* the pin and names no series above it, so a new quarter takes the version whole, year boundary included.

One thing the issue didn't cover, found while verifying against the mirrors: **Kali ships a respin as `kali-2025.1c/` in place of the `kali-2025.1/` it replaces** (`old.kali.org/kali-images/` lists `2025.1a` and `2025.1c` with no plain `2025.1`), and the letter rides into the ISO filenames inside. That is exactly the retire-and-replace case this issue is about, so the version grammar accepts a trailing respin letter; `isNewerFilename`'s existing string tiebreak already orders `2025.1c` above `2025.1a` above `2025.1`.

## Changes

- **`versionedSegment`** (new) — splits a path segment into the literal naming it and the version it ends with: `12.15.0` → `("", "12.15.0")`, `kali-2026.2` → `("kali-", "2026.2")`. Rejecting anything but digits-then-optional-letters in the last component is what keeps Debian's `12.15.0-live/` and `12.3.0-NEVER-RELEASED/` siblings, plus `kali-weekly/` and `current/`, out of a bump.
- **`pinnedVersionDirectory`** — returns that literal alongside the version, and candidates must carry the same one.
- **`fixedComponentCount`** (new) — derives how many leading components a bump must preserve from the entry's version label, per the rule above. A label naming no part of the pin fixes the major, the conservative reading.
- **`newerPinnedDirectory` → `newerPinnedEntry`** — returns the whole rolled-forward entry (bumped URL plus every field restating the old pin) and its notes, rather than just a URL.
- The comment at the version-label check said labels are "reported rather than rewritten"; that is now true of a label naming something *other* than the pin, and says so.

## Test plan

- [x] `swift Tools/regen-linux-image-catalog.swift` — 9/9 entries resolve, output byte-identical to the committed catalog
- [x] Live run with Kali staled to `kali-2026.1/` and Debian to `12.14.0/`: Kali rolled to `kali-2026.2/` with its label (`2026.1` → `2026.2`) and `isoPattern` rolled with it and the ISO resolved; Debian rolled to `12.15.0/` with label and pattern untouched, **not** to the 13.x.y sitting beside it
- [x] Live run with Kali pinned at the archived respin `kali-2025.1a/`: bumped across the year boundary to `kali-2026.2/`. That host rate-limits back-to-back requests so the manifest fetch 429s — confirmed as the host's limiter, not the bump, by probing the same URL standalone (200) and then index-then-manifest back to back (200 then 429)
- [x] Offline checks over the script's own helpers (extracted, not duplicated): segment parsing and every rejection above, both pin shapes plus the three unpinned catalog URLs, fixed-component derivation, respin and point-release ordering, and candidate selection over the real Debian and Kali index listings — 40/40
- [x] `make lint` — clean
- [x] `make test` — 2176 cases, `** TEST SUCCEEDED **`

Closes #728

https://claude.ai/code/session_01BhVWVjEm9w6MR7Lv41FHNx

### DIFF
--- a/Tools/regen-linux-image-catalog.swift
+++ b/Tools/regen-linux-image-catalog.swift
@@ -15,7 +15,8 @@
 // Which distributions are offered is a curated decision, so the entry list is
 // edited by hand and this script never adds or removes one. What it refreshes
 // is what the mirrors change on their own: the byte size behind each entry,
-// and the version-pinned archive directory an oldstable entry points at.
+// and the version-pinned directory an entry points at, along with every field
+// of that entry restating the version the pin names.
 //
 // ## How an entry resolves
 //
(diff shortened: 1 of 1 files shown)


nicholas-lonsinger:
## Review fixes

Two low findings from `/code-review medium`, both fixed in 296e701. The two Codex reviews (standard and adversarial) returned no findings.

**1. `fixedComponentCount` rolled a bare version pin whole.** A bump took the entire version whenever the entry's label restated the pin. That is right behind a literal, but a bare version directory is how both Debian's archive and Ubuntu's `releases/` lay out *every* release of the distribution — so an entry there naming its pin whole (say `releases/24.04/` labelled `"24.04 LTS"`) would have rolled onto 26.04 with directory, label and pattern rewritten together, succeeding while silently changing what the entry offers. The whole-version reading is now taken only behind a literal.

I took the suggested gate but not its reasoning. A literal is not *why* Kali rolls freely — both layouts list other releases of the same distribution side by side. The real reason is ambiguity: a label matching a bare version directory says nothing about whether the entry tracks the newest release or is pinned to that one, so the conservative reading is to hold. The doc now says that rather than asserting a mechanism that isn't there.

Worth keeping in view: the review's observation that Debian "fails safe here only by accident" is correct — its `isoPattern` retains the literal `debian-12.`, so a wrong roll breaks the glob and errors out rather than shipping. That accident is no longer what protects the case.

**2. A failed entry discarded the notes describing its own bump.** Notes reached the report only on the success path, so a bump that then failed named a directory — and, after this branch, an `isoPattern` — appearing nowhere in `LinuxImageCatalog.json`, with nothing saying the script synthesized them. I hit this during verification and had to infer the bump from the URL. Each failure now carries its notes:

```
•   kali-current — no checksum manifest at https://cdimage.kali.org/kali-2026.2/SHA256SUMS-does-not-exist — a respin renames the manifest, so check the directory and update checksumManifest
•       directory bumped from https://cdimage.kali.org/kali-2026.1/
•       version '2026.1' → '2026.2'
•       isoPattern 'kali-linux-2026.1*-installer-arm64.iso' → 'kali-linux-2026.2*-installer-arm64.iso'
```

An entry that fails without a bump prints none, as before.

### Re-verification

- `swift Tools/regen-linux-image-catalog.swift` — 9/9 resolve, output byte-identical to the committed catalog
- The archive-host respin run that previously 429'd now completes: `kali-2025.1a/` → `kali-2026.2/` across the year boundary, resolving the ISO inside with label and pattern rolled with it
- Offline checks 42/42, adding a bare pin whose label restates it whole holding against the next LTS listed beside it
- `make lint` clean; `make test` 2176 cases, `** TEST SUCCEEDED **`

Issue #727, opened 2026-08-06
Harden the Linux ISO destination against name collisions in ~/Downloads


nicholas-lonsinger:
## Found during
Security review of PR #725 (Linux distribution download catalog).

## Description
The download destination filename comes from the resolved manifest entry (mirror-influenced), and two disposal paths act on whatever file sits at that name in `~/Downloads`: a digest mismatch trashes it (recoverable), and `DownloadBundle.finalize` permanently removes a file that appears at the destination between the skip-existing check and finalization. A mirror that names an image after a file the user already has can therefore cause that file to be trashed (after failing verification) or, in the mid-download window, permanently replaced. Bounded to `*.iso` names in one directory and mostly recoverable via the Trash, but the remote-chosen-name primitive is new with this pipeline — the macOS path's destination names are app-derived.

## Location
`VMLifecycleCoordinator.linuxDownloadDestination`, `VMLifecycleCoordinator.discardUnverifiedImage`, `DownloadBundle.finalize`.

## Suggested fix
Design when picked up; candidates noticed at filing include a per-entry destination discriminator (the URL-digest suffix pattern `RestoreImageFilename.unique(for:)` already ships) or refusing the fast path when the existing file predates the download attempt.

nicholas-lonsinger:
From a follow-up review of `2e17497`, a fact that constrains this fix: on the Linux path there is no consented-replace route at all, because `LinuxInstallContext.requestedFreshDownload` is never set to `true`.

`VMCreationViewModel.buildLinuxInstallContext` always constructs `LinuxInstallContext(entry:)` and takes the initializer's `false` default — unlike `buildInstallContext`, which forwards the wizard's `requestedFreshDownload` on all three macOS download sources. The only other production writer is `VMLifecycleCoordinator.downloadLinuxImage`, which clears the flag. So the honored-once block in `downloadLinuxImage`, including its non-ISO destination refusal and the `invalidDownloadDestination` throw, is unreachable in the shipped flow; only `VMLifecycleCoordinatorTests` and `LinuxInstallContextTests` construct it `true`.

That is why the trashing described above happens with no consent step: the macOS path gates its equivalent behind the wizard's `shouldShowOverwriteWarning` / `confirmedOverwritePath` pair, and the Linux path has nothing to gate on, since the filename isn't known until resolve time.

Whichever way this is fixed, the flag is part of it — wired to a real prompt, or deleted along with the dead honored-once block.

PR #743, opened 2026-08-06
feat: Add a custom download URL as a third Linux installer-image source

### LINKED ISSUE #734: A catalog distribution picked under EFI persists onto a Linux-kernel-boot VM
## Found during

Review of `2e17497` (Linux distribution download catalog, PR #725).

## Description

`buildLinuxInstallContext()` returns the picked catalog entry regardless of the wizard's boot mode, so a distribution chosen while the Linux boot control reads **EFI** survives a switch to **Linux Kernel** and is persisted onto the created VM.

Verified in code:

- `buildLinuxInstallContext()` is the only boot-mode-sensitive value the wizard produces that isn't gated on `selectedBootMode == .efi`. In `buildConfiguration()` the installer `storageDisks` branch is gated, and so are `kernelPath`, `initrdPath`, `kernelCommandLine` and both kernel bookmarks.
- `BootConfigContentViewController.bootModeChanged` rebuilds the conditional UI but does not clear `linuxSelection`.
- `bootConfigValid` accepts `.linuxKernel` on `kernelPath != nil` alone, and the Review step switches on boot mode and renders only the kernel rows — so nothing in the wizard surfaces the stale pick.
- `VMLibraryViewModel.createVM` assigns `config.linuxInstallContext = wizard.buildLinuxInstallContext()` for every `.linux` guest, with no boot-mode condition.

The created VM therefore opens in the initial-boot state with the download banner and Download button (`GuestSetupDescriptor` and `DetailBannerView` both key off `linuxInstallContext` alone), and the first Start resolves, downloads the ISO, and prepends it as `storageDisks[0]` on a VM that boots from the kernel.

Reproduction: create a Linux VM, leave the boot control on EFI, choose a distribution from the catalog, switch the control to Linux Kernel, pick a kernel, finish the wizard, Start.

Coverage gap: `buildLinuxInstallContextMatchesBuildConfiguration` exercises `.efi` only, and no test drives a boot-mode switch after a catalog pick.

## Location

`VMCreationViewModel.buildLinuxInstallContext`, `VMCreationViewModel.bootConfigValid`, `BootConfigContentViewController.bootModeChanged`, `VMLibraryViewModel.createVM`.

## Suggested fix

Gate the context on the EFI boot mode the way its siblings in `buildConfiguration()` are, and/or clear `linuxSelection` when the boot mode changes. Design when picked up.

### LINKED ISSUE #739: Add a custom download URL as a third Linux installer-image source
## Summary

The creation wizard's macOS path offers a **custom download URL** installer source;
the Linux path does not. Add a third image source to the Linux boot step so a user can
paste a URL for an installer ISO — with an optional SHA-256 — alongside the bundled
catalog and a local ISO.

## Context

**macOS sources** — `IPSWSource` has four cases: `.downloadLatest`, `.catalogVersion`,
`.customURL`, `.localFile`. `IPSWSelection.customURL` carries a `ProbedRestoreImage`,
produced by `RestoreImageProbeService.probe` through the nested
`RestoreImageURLSheetContentViewController`. The sheet's **Use** button stays disabled
until the probe succeeds; the probe reads the zip end-of-central-directory via ranged
GETs to confirm the image carries the VM hardware model, rather than fetching the file.
A pasted URL then shares the download destination, existing-file overwrite warning, and
resume affordance with the two catalog-backed download sources
(`IPSWSource.downloadsImage` is `true` for all three).

**Linux sources** — `LinuxImageSelection` has two cases, `.catalogEntry` and
`.localISO`, rendered as the two radios in `BootConfigContentViewController`'s private
`ImageSource` enum under the EFI boot mode. (The Linux-kernel boot mode has no image
source picker at all.) There is no equivalent of `.customURL`. A distribution absent
from `Kernova/Resources/LinuxImageCatalog.json` can only be installed by downloading
the ISO in a browser first and picking it as a local file.

**Verification differs between the two paths, and Linux is the stricter one.**
`LinuxImageResolveService` turns a catalog entry into a `ResolvedLinuxImage` whose
`sha256` is non-optional, sourced from the distribution's published checksum manifest;
`VMLifecycleCoordinator.downloadLinuxImage` hashes the downloaded file with
`FileDigest.sha256`, and a mismatch throws and trashes the file before anything is
attached. The macOS path runs no digest comparison at any point — its assurance is the
wizard-time structural probe plus `VZMacOSRestoreImage`'s own validation at install
time. A user-supplied Linux URL has no manifest behind it, so a URL pick can only be
verified against a digest the user supplies.

**A catalog entry is deliberately not a fixed URL.** An entry names a directory, an ISO
filename glob, and a checksum manifest, so a point release that renames the ISO in place
does not break the catalog. A pasted URL is a fixed URL and carries the opposite
property.

Also relevant to whichever shape is chosen: `buildLinuxInstallContext()` returns `nil`
for a local-ISO pick, so `LinuxInstallContext` exists only for download-based sources,
and it carries a non-optional `LinuxImageCatalogEntry` today. `bootConfigValid` gates
Linux EFI on `linuxSelection != nil`. Neither `LinuxInstallContext` nor
`ResolvedLinuxImage` can currently represent an image with no published digest.

Wiring and API design are deliberately omitted here; do that when the work is picked up.

## Requirements

- Under the EFI boot mode, the Linux boot step offers a third image source that takes a
  user-supplied URL, alongside "Choose a Distribution…" and "ISO File…".
- The URL entry carries an **optional SHA-256 field**. When filled, the existing verify
  step runs the downloaded file against it — `FileDigest.sha256`, mismatch trashes the
  file — the same treatment a catalog pick's manifest digest gets. When empty, the
  download is not verified.
- **The scheme requirement follows the digest.** With no digest supplied, the URL must
  be HTTPS. With a digest, HTTP is accepted, since the digest carries integrity
  independently of the transport. (`RemoteFileSizeProbe` and `RestoreImageProbeService`
  refuse non-HTTPS unconditionally today.)
- **When no digest is entered, the UI says the download will not be verified and that
  the user should choose a host they trust.** The note states that and nothing more — no
  invented consequence.
- A pasted digest is validated for shape before it is accepted (64 hex characters,
  case-insensitive), so a malformed paste fails at entry rather than after a multi-GB
  download.
- The URL itself is checked before anything is downloaded, and the confirm control stays
  disabled until it passes — matching how the macOS URL sheet gates **Use**. What
  constitutes a passing check for an ISO needs a decision: the macOS probe's IPSW zip
  inspection has no direct ISO analog, so the check may reduce to the well-formed-URL
  and scheme refusals plus the reachable, non-zero size `RemoteFileSizeProbe` already
  provides.
- A URL pick reaches the created VM the way a catalog pick does: the initial-boot banner
  and Download button appear, and the first Start downloads, verifies (when a digest was
  supplied), and attaches the ISO, with the shared `GuestSetupProgressViewController`
  reporting progress. `GuestSetupState.linuxImage()`'s Verify step reflects whether
  there is anything to verify.
- The Review step shows what a URL pick will install, including whether it will be
  verified.
- The download destination, the existing-file overwrite warning, and the resume
  affordance behave as they do for a catalog pick. Note that
  `RestoreImageFilename.unique(for:)` already derives a collision-free destination name
  for off-convention macOS URLs; the Linux destination has no such discriminator today
  (#727).
- A URL pick must not persist onto a VM switched to Linux-kernel boot — the same
  constraint #734 raises for catalog picks.
- `LinuxInstallContext` is a persisted `Codable`; per AGENTS.md the format is
  current-only, so any reshaping ships without migration code.
- Tests cover the new source and its failure paths — including a verified pick, an
  unverified pick, a digest mismatch, a malformed digest, and an HTTP URL with and
  without a digest — mirroring the coverage the macOS `.customURL` path has across
  `RestoreImageURLSheetContentViewControllerTests`,
  `IPSWSelectionContentViewControllerTests` and `VMCreationViewModelTests`, and
  extending `BootConfigContentViewControllerTests` and `VMLifecycleCoordinatorTests`.

## Dependencies

- Related to #727 — a user-supplied URL widens the remote-chosen-filename surface in
  `~/Downloads` that issue is about.
- Related to #726 — both concern how much trust the Linux image pipeline places in the
  source it fetched from.
- Related to #734 — the boot-mode gating that issue asks for applies to this source too.

## Additional Notes

- A supplied digest is only as good as where the user got it. The check this feature
  adds is that the bytes on disk match the digest that was typed in; it says nothing
  about where that digest came from.
- No "download latest" analog is proposed: "latest" is per-distribution, and catalog
  entries already resolve to the current file at pick time.

nicholas-lonsinger:
## Summary

The creation wizard's Linux boot step now offers **Image URL…** under EFI, beside "Choose a Distribution…" and "ISO File…". It takes a URL and an **optional SHA-256**.

- **Verification follows the digest.** With one supplied, the existing verify step runs `FileDigest.sha256` against it and a mismatch trashes the file — the same treatment a catalog pick's manifest digest gets. With none, the download isn't verified, `GuestSetupState.linuxImage(hasVerifyStep:)` draws no Verify step, and the sheet, the boot-step badge and the Review row all say so.
- **The scheme follows the digest.** No digest → HTTPS required. Digest → HTTP accepted, since the digest carries integrity independently of the transport. `RemoteFileSizeProbe.size(of:allowingHTTP:)` is the one place that relaxes, for the one caller that holds a digest.
- **The check is the same code the download runs.** `LinuxImageResolving` gains a `resolve(_ image: CustomLinuxImage)` overload, and both the wizard's pre-download check and the install pipeline's per-attempt resolve go through it. There is one set of admission rules — digest shape, scheme, and the `.iso` filename the destination is named from — enforced at entry and again at use time, because the persisted value comes off a `config.json` a user can edit.

## What the URL check establishes

An ISO has no structure analogous to an IPSW's zip central directory, so there is no equivalent of `RestoreImageProbeService`'s installability probe. The check reduces to what the issue anticipated: a well-formed URL naming a host, an admissible scheme, a last path component that sanitizes to an `.iso` filename, and a reachable non-zero size from `RemoteFileSizeProbe`. **Use** stays disabled until all of it passes, matching how the macOS URL sheet gates its own **Use**.

Requiring the URL to name an `.iso` is the one constraint beyond the issue's minimum. It buys three things: the destination filename is something the user can see in the link they pasted, the `.iso` extension the fresh-download guard in `downloadLinuxImage` already demands is guaranteed, and no second copy of `RestoreImageFilename`'s digest-discriminator machinery is needed. A redirect-style download link is refused at entry with an actionable message rather than landing on a generated name.

## Persisted shape

`LinuxInstallContext` carried a non-optional `LinuxImageCatalogEntry`, which cannot represent a URL pick. It now carries a `Source` enum:

```swift
enum Source: Sendable, Equatable {
    case catalogEntry(LinuxImageCatalogEntry)
    case customURL(CustomLinuxImage)
}
```

`Codable` is hand-written rather than synthesized so the on-disk document stays flat (`"source": "customURL"`, `"remoteURL"`, `"sha256"`) instead of nesting under a synthesized case name — the shape `MacOSInstallContext` already writes. Unlike that type, the in-memory value cannot hold a source without its payload, so the coordinator's branch is exhaustive rather than a guard that throws.

Per AGENTS.md the format is current-only, so **no migration code ships**. A VM created before this change whose Linux download is still pending will fail to decode and be reported as an unloadable bundle.

## Also fixed

`buildLinuxInstallContext()` is now gated on `selectedBootMode == .efi`, alongside the installer `storageDisks` and every kernel field in `buildConfiguration()`. That is the fix #734 asks for, and the constraint this issue restates for the new source — one gate covers both.

## Rejected alternatives

- **A separate wizard-only probe service.** The wizard-time check and the download-time resolve would have been two implementations of one rule set, which is exactly how a scheme or filename refusal drifts between them. One overload on the existing seam instead.
- **Persisting the size read at wizard time.** `expectedSizeBytes` is a hard ceiling on the transfer, and a size read days before the Start would abort a legitimate download. The URL is re-resolved on every attempt for a fresh size — the same "resolved on every attempt" rule the catalog path follows, for a different reason.
- **A wizard-time overwrite warning.** A Linux catalog pick has none, because its destination isn't known until the mirror is asked. The requirement is parity with a catalog pick, so nothing new was invented here.
- **Flat optional fields on the context**, matching `MacOSInstallContext`. Type-safety in the code that branches on the source won out; the on-disk shape is the same either way.

## Test plan

- [x] `make build`
- [x] `make test` — full suite green
- [x] `make lint`
- [ ] Wizard: pick **Image URL…**, paste an HTTPS `.iso` link with and without a checksum; confirm **Use** gates on the check, the unverified banner appears, and the Review step shows the file, size, verification and destination
- [ ] Paste an `http://` link with and without a checksum; confirm the refusal and the acceptance
- [ ] Paste a malformed checksum; confirm it fails at entry with no network call
- [ ] Create from a URL pick and Start: the ISO downloads, verifies when a digest was given, and attaches as the boot disk

New coverage: `CustomLinuxImageTests` and `LinuxImageURLSheetContentViewControllerTests`, plus additions to the resolve, boot-config, review, creation-model, install-context, setup-state, descriptor and lifecycle suites — a verified pick, an unverified pick, a digest mismatch, a malformed digest, and HTTP with and without a digest.

## Notes

- No `RATIONALE:` comments added.
- `ChecksumManifest.isSHA256` is no longer private, so a manifest's digest and a user's are admitted on one rule.
- `attachVerifiedImage` is renamed `attachInstallerImage`: it now runs for an image with nothing to verify against.
- #727 (destination name collisions in `~/Downloads`) is untouched and still open. A URL pick's destination is derived the same way a catalog pick's is, so it neither widens nor narrows that surface beyond what the issue already describes.

Closes #739
Closes #734

https://claude.ai/code/session_01LmhDjnAbqPdet79K6hFqrS

### DIFF
--- a/Kernova/Models/ChecksumManifest.swift
+++ b/Kernova/Models/ChecksumManifest.swift
@@ -71,8 +71,9 @@ enum ChecksumManifest {
     /// Whether `candidate` is a SHA-256 digest written as 64 hex digits.
     ///
     /// A SHA-512 line, which several of these mirrors carry beside the SHA-256
-    /// one, fails on the length.
-    private static func isSHA256(_ candidate: String) -> Bool {
+    /// one, fails on the length. Also the shape a digest typed into the wizard
+    /// is admitted on, so a manifest's digest and a user's are held to one rule.
+    static func isSHA256(_ candidate: String) -> Bool {
         candidate.count == 64 && candidate.allSatisfy(isHexDigit)
     }
 
--- a/Kernova/Models/CustomLinuxImage.swift
+++ b/Kernova/Models/CustomLinuxImage.swift
@@ -15,15 +15,28 @@ struct CustomLinuxImage: Codable, Sendable, Equatable {
     /// verified.
     var sha256: String?
 
+    /// The name ``url`` itself gives the image, or a stand-in when it gives
+    /// none.
+    ///
+    /// Shown, never written: nothing on disk is named from this, so it does not
+    /// have to be safe to append to a directory and does not throw.
+    var displayName: String {
+        SafeFilename.sanitized(url.lastPathComponent, requiring: "iso") ?? "the installer image"
+    }
+
     /// The filename this image's download lands on, after every condition
     /// ``make(urlText:checksumText:)`` checked is checked again.
     ///
     /// Re-checked at use time because the value reaching here comes off a
     /// `config.json` a user can edit — the same reason `LinuxImageResolveService`
-    /// re-admits a catalog entry before contacting its mirror. The scheme
-    /// requirement follows the digest: without one, nothing but the transport
-    /// stands behind the bytes.
-    func validatedFilename() throws -> String {
+    /// re-admits a catalog entry before contacting its mirror.
+    ///
+    /// The name is unique to ``url`` rather than taken from it. A link ending
+    /// in a name the user already has in Downloads would otherwise resolve to
+    /// that file, which `DownloadService` adopts as the download in place of
+    /// fetching anything — installing an image the user never chose when there
+    /// is no digest, and trashing their file when there is one.
+    func destinationFilename() throws -> String {
         if let sha256, !ChecksumManifest.isSHA256(sha256) {
             throw LinuxImageURLError.malformedChecksum
         }
@@ -32,20 +45,28 @@ struct CustomLinuxImage: Codable, Sendable, Equatable {
         guard let scheme = url.scheme?.lowercased() else {
             throw LinuxImageURLError.malformedURL
         }
-        guard scheme == "https" || scheme == "http" else {
-            throw LinuxImageURLError.unsupportedScheme
+        // HTTPS only, whatever digest is supplied: App Transport Security
+        // refuses a cleartext load to any public host before the request is
+        // issued, so admitting one here would only defer the refusal into an
+        // error that says nothing the user can act on.
+        guard scheme == "https" else {
+            throw LinuxImageURLError.insecureURL
         }
         guard url.host() != nil else {
             throw LinuxImageURLError.malformedURL
         }
-        guard scheme == "https" || sha256 != nil else {
-            throw LinuxImageURLError.insecureURL
-        }
-        guard let filename = SafeFilename.sanitized(url.lastPathComponent, requiring: "iso") else {
+        // A direct link to a named `.iso` is what makes the generated name
+        // recognizable and the destination an ISO, which is what the
+        // replace-existing guard in the download pipeline keys on.
+        guard SafeFilename.sanitized(url.lastPathComponent, requiring: "iso") != nil else {
             throw LinuxImageURLError.notAnISOLink
         }
-        return filename
+        return UniqueDownloadFilename.make(
+            for: url, fileExtension: "iso", defaultStem: Self.defaultStem)
     }
+
+    /// Stem of a generated filename when the URL's own last component yields none.
+    private static let defaultStem = "LinuxImage"
 
     /// The image `urlText` and `checksumText` name, or the first reason they
     /// name none.
@@ -57,7 +78,7 @@ struct CustomLinuxImage: Codable, Sendable, Equatable {
         guard let url = URL(string: urlText.trimmingCharacters(in: .whitespacesAndNewlines))
         else { throw LinuxImageURLError.malformedURL }
         let image = CustomLinuxImage(url: url, sha256: sha256)
-        _ = try image.validatedFilename()
+        _ = try image.destinationFilename()
         return image
     }
 
--- a/Kernova/Models/CustomLinuxImage.swift
+++ b/Kernova/Models/CustomLinuxImage.swift
@@ -1,0 +1,74 @@
+import Foundation
+
+/// An installer ISO at a user-supplied URL: where to fetch it, and the digest
+/// the user typed in to check it against.
+///
+/// ``sha256`` is what the download is verified against, exactly as a catalog
+/// pick is verified against its mirror's manifest. It says nothing about where
+/// the user got the digest — only that the bytes on disk match it.
+struct CustomLinuxImage: Codable, Sendable, Equatable {
+    /// Where the ISO is served from.
+    var url: URL
+
+    /// The digest to check the download against, as 64 lowercase hex
+    /// characters, or `nil` when the user supplied none and the download is not
+    /// verified.
+    var sha256: String?
+
+    /// The filename this image's download lands on, after every condition
+    /// ``make(urlText:checksumText:)`` checked is checked again.
+    ///
+    /// Re-checked at use time because the value reaching here comes off a
+    /// `config.json` a user can edit — the same reason `LinuxImageResolveService`
+    /// re-admits a catalog entry before contacting its mirror. The scheme
+    /// requirement follows the digest: without one, nothing but the transport
+    /// stands behind the bytes.
+    func validatedFilename() throws -> String {
+        if let sha256, !ChecksumManifest.isSHA256(sha256) {
+            throw LinuxImageURLError.malformedChecksum
+        }
+        // Scheme before host, so a `file:` or `ftp:` link — neither of which
+        // carries one — is refused for what it is rather than as unparseable.
+        guard let scheme = url.scheme?.lowercased() else {
+            throw LinuxImageURLError.malformedURL
+        }
+        guard scheme == "https" || scheme == "http" else {
+            throw LinuxImageURLError.unsupportedScheme
+        }
+        guard url.host() != nil else {
+            throw LinuxImageURLError.malformedURL
+        }
+        guard scheme == "https" || sha256 != nil else {
+            throw LinuxImageURLError.insecureURL
+        }
+        guard let filename = SafeFilename.sanitized(url.lastPathComponent, requiring: "iso") else {
+            throw LinuxImageURLError.notAnISOLink
+        }
+        return filename
+    }
+
+    /// The image `urlText` and `checksumText` name, or the first reason they
+    /// name none.
+    ///
+    /// Both arrive as typed, whitespace included. An empty `checksumText` is a
+    /// deliberate "don't verify this", not a malformed digest.
+    static func make(urlText: String, checksumText: String) throws -> CustomLinuxImage {
+        let sha256 = try normalizedChecksum(checksumText)
+        guard let url = URL(string: urlText.trimmingCharacters(in: .whitespacesAndNewlines))
+        else { throw LinuxImageURLError.malformedURL }
+        let image = CustomLinuxImage(url: url, sha256: sha256)
+        _ = try image.validatedFilename()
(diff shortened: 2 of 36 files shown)


nicholas-lonsinger:
## Review round: three passes, two substantive findings, both fixed in `b0914da`

Ran `/code-review medium`, a Codex review, and a Codex adversarial review independently. They converged on the same two issues — the collision was the top finding in all three.

### 1. Destination collision — *fixed*

`/code-review` medium · Codex **P1** · adversarial **high**

`buildLinuxInstallContext()` pinned the destination to `~/Downloads/<name-from-the-URL>`, and `DownloadService` (`DownloadService.swift:114`) treats any completed file already at the destination as the download and skips the transfer. Paste a link ending in a name the user already has and:

- **no checksum** → that unrelated ISO is attached as boot media and the install context cleared, with nothing fetched. The only path in the app where an unchecked, un-downloaded file becomes the installer.
- **with a checksum** → the digest mismatch trashes the user's pre-existing file.

Catalog picks are insulated (mirror-specific names, always a digest). Generic names in pasted links make this far likelier here.

**Fix:** the destination is now unique to the URL rather than taken from it — the `RestoreImageFilename.unique(for:)` discriminator this issue itself pointed at, extracted into `UniqueDownloadFilename` so both sources share one implementation. Nothing unrelated can occupy a URL-unique path, which closes both halves. One URL always maps to one file, so resume is unaffected.

Because the on-disk name now carries a suffix the user never typed, `CustomLinuxImage` splits the two concepts: `displayName` (the name in the link — badge, Review "File", the initial-boot banner) and `destinationFilename()` (what lands — "Save to", the download, the disk label). `LinuxImageSelection.customURL` carries the `CustomLinuxImage` plus the checked length, which also removed the round-trip reconstruction in `buildLinuxInstallContext()`.

This does **not** close #727 — that is the same class of exposure for a catalog pick's mirror-named destination.

### 2. HTTP-with-checksum is unreachable — *fixed*

`/code-review` low · Codex P2 · adversarial medium

ATS blocks cleartext loads to public hostnames before the request is issued, and there is no `NSAppTransportSecurity` key in the tree. Every digest-backed `http://` link would pass admission and then fail with an opaque ATS error; the allowance only ever worked for IP-literal and `.local` hosts.

**Fix: HTTPS only**, whatever checksum is supplied. `RemoteFileSizeProbe.size(of:allowingHTTP:)` reverts to HTTPS-only, and `unsupportedScheme` folds into `insecureURL`. `NSAllowsArbitraryLoads` was rejected — an app-wide cleartext permission plus an App Store review justification for one narrow path.

This contradicts #739's stated scheme requirement, so [the issue is annotated](https://github.com/nicholas-lonsinger/kernova/issues/739#issuecomment-5201323951) with the reasoning.

### 3. Two comments this branch falsified — *fixed*

`VMInstance+Display.swift` claimed a Linux context has no destination until its first resolution (a URL pick now carries one from creation), and `BootConfigContentViewController` still said "both sources" in a three-way switch.

### Dismissed

Nothing. All three findings cleared the severity bar; the passes raised no style-only or hypothetical findings worth recording.

`make build`, `make test` (exit 0) and `make lint` green locally.

Issue #487, opened 2026-07-09
Sequential batch import widens the window for cross-invocation hasPreparing collisions


nicholas-lonsinger:
## Found during

Multi-agent /code-review of PR #483 (fix #444 — sequential-await multi-bundle import).

## Description

PR #483 fixes multi-select import (all bundles in *one* batch call now import, since each is awaited before the next starts) by keeping the existing global `hasPreparing` guard exactly as-is and just having `importVMs(from:)` await sequentially within its own loop. That guard is still a single, cross-invocation flag (`instances.contains(where: \.isPreparing)`), so two *independent* import triggers (e.g. a drag-and-drop batch still mid-copy on bundle 3 of 5, and a separate Finder double-click of one more `.kernova` bundle arriving via `application(_:open:)`) can still collide: the second trigger's `importVM` sees `hasPreparing == true` and rejects with `PreparingError.operationInProgress` — the same "operation in progress" alert #444 was filed to eliminate, just from two overlapping user actions instead of one multi-select action.

This isn't a new failure mode (pre-#444, only one bundle was ever actually mid-copy at a time too, and a second concurrent trigger would have hit the same guard) — but the *window* during which it can occur is now wider: a batch now holds `hasPreparing == true` for the sum of all its bundles' copy durations, rather than being bounded by a single file's copy time (since before the fix, a batch's own later bundles rejected near-instantly rather than actually copying). The PR's own design rationale argued the guard's cross-operation behavior is "unchanged from today," which undersells this widened-window effect.

## Location

`VMLibraryViewModel.hasPreparing` (the global cross-invocation flag), `VMLibraryViewModel.importVM(from:)`'s `guard !hasPreparing` gate, `VMLibraryViewModel.importVMs(from:)` (the sequential loop that widens the flag's true-duration).

## Suggested fix

A FIFO import queue (the alternative PR #483 explicitly considered and rejected as unnecessary machinery for the *within-one-batch* case) would also close this cross-invocation gap, since it would queue rather than reject a second trigger's bundles. Given the tradeoff was already litigated for #444 and the queue was judged not worth the added complexity for that narrower bug, revisit whether the wider cross-invocation exposure changes that calculus, or whether this residual gap is acceptable given it only affects two independent user-triggered imports actually overlapping in time (a comparatively rare edge case). Design when picked up.

PR #794, opened 2026-08-11
refactor: Load the VM library after launch, off the main actor

### LINKED ISSUE #792: Load the VM library after launch instead of before applicationDidFinishLaunching
## Found during
The 2026-08-10 switcher-order investigation (PR #790, docs/research/2026-08-10-xcode-launch-switcher-tail.md).

## Description
The VM library loads synchronously before the app finishes launching: `AppDelegate.main()` constructs `AppDelegate`, whose `init` constructs `VMLibraryViewModel`, which loads the full library — all before `app.run()`. On 2026-08-10 (debug build, 10-VM library) "Loaded 10 VMs" logged ~1.9 s after process start, and cold-launch resolution plus the first window followed it, ~2.2 s after start. Every manual launch pays that delay before anything appears on screen, and the same delay is what pushed the cold-launch `.accessory` → `.regular` morph seconds past launch — the failing condition in the research note above.

Expected shape: finish launching and present the window promptly; the library loads asynchronously into it (loading state until populated).

A fast `applicationDidFinishLaunching` is also a precondition if the launch-posture inversion discussed on PR #790 (start `.regular`, demote only on login-item provenance) is pursued — the demote window at login is bounded by launch speed.

## Location
`AppDelegate.main()`, `AppDelegate.init` (the `viewModel` property), `VMLibraryViewModel` initialization.

## Suggested fix
Defer library loading until after `applicationDidFinishLaunching`; construct the view model empty and load asynchronously, with the window showing immediately. Design when picked up.

nicholas-lonsinger:
Closes #792

## Summary

The VM library was read inside `VMLibraryViewModel.init`, which `AppDelegate.init` runs *before* `NSApplication.run()`. Every launch paid the whole read — a directory listing plus a JSON decode and a stat per bundle — on the main thread, before the app could finish launching or put a window on screen.

The read now starts from `applicationWillFinishLaunching` and does its file I/O off the main actor. Nothing on the launch path touches disk any more; the window is presented and the library fills in behind it.

## Changes

**`VMLibraryViewModel`**
- `init` no longer loads, and no longer starts the directory watcher — only the sleep watcher, which registers notification observers and does no I/O.
- New `startLibrary()` loads and *then* starts the directory watcher. The order is load-bearing: the watcher's `reconcileWithDisk()` re-reads every bundle synchronously on the main actor, so starting it first would let it both race the initial load and reintroduce the stall this change removes.
- `loadVMs()` is now `async`, split along the isolation boundary it always implicitly had:
  - `scanLibrary(using:)` — `nonisolated`, runs in a detached task, touches no view-model state, returns a `Sendable` `LibraryScan`.
  - `apply(_:)` — `@MainActor`, builds the `VMInstance`s (which are `@MainActor`) and applies order and selection.
- `hasLoadedLibrary` records that the first read finished, including when it failed: a failed listing answers "no VMs" too, and UI must not wait on a load that is over.

**`AppDelegate`** — owns the load `Task`, started in `applicationWillFinishLaunching` — ahead of `application(_:open:)`, which waits on it. `resolveColdLaunch(from:)` is untouched and still runs while the launch Apple event is current.

**`DetailContainerViewController`** — shows a bare pane until the library has been read, and tracks `hasLoadedLibrary` in its observation.

## The empty-then-populated interval

The one place the transition is user-visible is the detail pane. Its empty state reads "No Virtual Machine Selected — Select a virtual machine from the sidebar or create a new one" and carries a **New Virtual Machine** button. Shown to someone with a full library while the read is in flight, that is both false and actionable, so the pane withholds it until the library is known. No spinner: SPEC.md asks for nothing here, and a blank pane for the duration of a local file read is the honest minimum.

The sidebar and the status item needed no changes — both are observation-driven, and the status menu is rebuilt on open.

## A correctness hazard this had to handle

`application(_:open:)` is delivered between `applicationWillFinishLaunching` and `applicationDidFinishLaunching`, while the read is still in flight, and `importVMs(fromDroppedURLs:)` dedups by UUID against `instances`. Against a not-yet-read library that dedup misses — so double-clicking a bundle that is *already in the library* (reachable via **Open VMs Folder**) would copy a multi-GB bundle a second time instead of selecting the existing VM. The open path now waits for the load.

The wait sits at that call site rather than inside `importVMs(fromDroppedURLs:)` because that method's reservation pass is deliberately suspension-free, so overlapping triggers see each other's phantoms — the property #487 fixed and its test pins.

## Deviation from the issue

The issue reports the load costing ~1.9 s of pre-window time. The unified log for that day does not support it: `Loaded N VMs` lands roughly 0.3-0.5 s after process start, and on the very launch the issue quotes, ~1.2 s sat *after* the load, inside `setupMainMenu()`'s first menu-bar/text-service materialization. The numbers in the issue line up with the `resident app ready` record, not the `Loaded N VMs` one.

The change is still worth making on its own terms — it takes all disk I/O out of the pre-`run()` path, which is a precondition for the launch-posture work discussed on #790, and it scales with library size rather than being fixed. But it is not a 1.9 s win, and anyone chasing the remaining launch latency should look at `setupMainMenu()` and status-item creation first.

## Test plan

- [x] `make build`, `make test`, `make lint` clean
- [x] New `VMLibraryViewModel` tests: `init` reads no storage; `hasLoadedLibrary` flips after an empty load and after a failed listing
- [x] New `DetailContainerLibraryLoadTests`: the empty state is withheld before the read, appears after a library that reads back empty (which pins the `hasLoadedLibrary` observation — nothing else changes in that case, so tracking only the instance list would strand the pane blank), and a loaded VM shows its detail content
- [x] Existing load/order/selection/reconcile tests updated to drive the load explicitly
- [ ] Behavioral launch verification by the maintainer

### DIFF
--- a/Kernova/App/AppDelegate.swift
+++ b/Kernova/App/AppDelegate.swift
@@ -16,7 +16,8 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSMenuItemValidation, 
     private let preferences: AppPreferences
     private var mainWindowController: MainWindowController?
     private let viewModel: VMLibraryViewModel
-    /// The library's first read from disk, started once launch is complete.
+    /// The library's first read from disk, started in
+    /// `applicationWillFinishLaunching`.
     ///
     /// Retained so `application(_:open:)` can wait for it: a Finder open that
     /// launched the app is delivered while the read is still in flight.
@@ -295,6 +296,18 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSMenuItemValidation, 
 
     // MARK: - NSApplicationDelegate
 
+    /// Starts the library read.
+    ///
+    /// Deliberately here and not in `applicationDidFinishLaunching`: AppKit
+    /// delivers a launch document's `application(_:open:)` *between* the two, and
+    /// that path waits on `libraryLoad` — left unset, the wait would silently
+    /// pass through and re-import a bundle already in the library. Starting the
+    /// read costs nothing here; the task body only runs once the main actor
+    /// yields, and its file I/O is off the main actor either way.
+    func applicationWillFinishLaunching(_ notification: Notification) {
+        libraryLoad = Task { @MainActor [viewModel] in await viewModel.startLibrary() }
+    }
+
     func applicationDidFinishLaunching(_ notification: Notification) {
         setupMainMenu()
 
@@ -324,12 +337,6 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSMenuItemValidation, 
         } else {
             startResidentApp()
         }
-
-        // Last, and asynchronously: the library read is disk-bound, and every
-        // step above — the cold-launch resolution most of all — has to reach the
-        // run loop without waiting on it. The window is already on screen and
-        // fills in when the read lands.
-        libraryLoad = Task { @MainActor [viewModel] in await viewModel.startLibrary() }
     }
 
     // MARK: - Resident App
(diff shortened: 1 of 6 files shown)


nicholas-lonsinger:
## Review pass

Three reviewers (built-in `/code-review`, Codex review, Codex adversarial) converged on one real defect, now fixed in 53a0600:

- **`apply(_:)` replaced `instances` wholesale**, discarding anything registered during the read. The case that bites is a preparing import phantom: its copy is uninterruptible, so losing the row left a multi-GB copy running against a bundle the library had forgotten, with no cancel affordance and `hasPreparing` back to `false` — the exact gap `reconcileWithDisk`'s `RATIONALE:` refuses to open. `loadVMs()` now snapshots the pre-read IDs and `apply` keeps anything newer than the read, letting it outrank the read's view of the same VM so no UUID is duplicated.
- **`application(_:open:)` awaited a `libraryLoad` that was still `nil`** — AppKit delivers a launch document *between* `applicationWillFinishLaunching` and `applicationDidFinishLaunching`, so the wait passed straight through and only the enclosing `Task`'s scheduling kept the re-import bug hidden. The load task is now created in `applicationWillFinishLaunching`.
- Stale `"loadVMs already ran during init"` test comments corrected.

Also added the interleaving test all three reviewers asked for, using a deterministic main-actor hand-off rather than a timing window.

Dismissed: the adversarial review's auto-start scenario — VMs are deliberately never auto-started (`startResidentApp`), so no discarded instance can hold a live VM.

## CI

The first run of 53a0600 failed on three `KernovaKitTests` supersession tests; re-running the identical commit passed. They are flakes in a target this diff never touches, and all three pass locally against the branch — filed as #795. CI is verified green on 53a0600.

Issue #796, opened 2026-08-11
Decide launch posture from a --login-launch argument instead of post-launch provenance


nicholas-lonsinger:
## Found during
The 2026-08-10 switcher-order investigation (PR #790, docs/research/2026-08-10-xcode-launch-switcher-tail.md) and the follow-up launch-posture design discussion.

## Description
Today every launch starts `.accessory` and manual launches morph to `.regular` after launch — the fragile transition behind the FB7743313 menu-bar defer, the cooperative-activation reassert, the debugger switcher-tail behavior in the research note above, and the `isDebuggerLaunched` stopgap in `AppDelegate.main()`. Launch provenance (login vs manual) is currently only readable after launch, from the `oapp` Apple event (FB10207829), which forces the pessimistic accessory-first default.

Decided design (maintainer, 2026-08-10) — provenance moves into argv so the posture decision is synchronous at process start:

- *Open at Login* registers `SMAppService.agent(plistName:)` — an embedded LaunchAgent plist (`BundleProgram` + `ProgramArguments` carrying `--login-launch`) — replacing `SMAppService.mainApp`.
- `main()` decides posture from argv alone: `--login-launch` present → `.accessory` before checkin (no Dock blip at any launch speed); absent → `.regular`, window shows.
- **No fallback code.** A user-added System Settings "Open at Login" entry (no argv) launches like a manual launch and opens the window — decided as acceptable.
- Deletions: the launch-event provenance inspection (`launchProvenance`, `keyAELaunchedAsLogInItem`, the FB10207829 dependency), `armColdLaunchActivationHeuristic` and its `RATIONALE:`, the `isDebuggerLaunched` guard (an Xcode Run is just a manual launch), and their tests.
- The login posture becomes exercisable from a terminal (`… /Kernova --login-launch`) instead of requiring a logout.
- Accepted trade-offs: System Settings presentation moves from the "Open at Login" list to the background-items section (attributed via `AssociatedBundleIdentifiers`); registration changes fire the BTM "background item added" notification.
- Not a dependency of or on #792 — the argv demote precedes checkin, so the login path has no blip regardless of load time. #792 stands on its own (manual-launch latency).
- The same argv pattern is a candidate for the TCC-revocation relaunch (restore prior headless/GUI state) — separate decision when picked up.

Open questions:
- Session restore (app running at logout, relaunched argv-less at login) now opens the window where today's heuristic stayed headless — confirm acceptable.
[...]

PR #803, opened 2026-08-12
fix: Register login launch through a bootstrap LoginItem helper

### LINKED ISSUE #801: Registering the app itself as the login LaunchAgent collides with its own process lifecycle
## Found during
Review of PR #800 (the `--login-launch` posture change, #796). Raised independently by four review passes.

## Description
PR #800 registers the app's own executable as the login LaunchAgent (`BundleProgram` = `Contents/MacOS/Kernova`, per the decided design in #796). Because the registered service *is* the long-running VM host, three behaviors documented in `SMAppService.h` on the macOS 27.0 SDK now apply to Kernova's own process.

**1. Turning *Open at Login* off kills the running app.** `unregisterAndReturnError` states: "If the service corresponds to a LoginItem, LaunchAgent, or LaunchDaemon and the service is currently running it will be killed. The unregister call will not wait for the service to be reaped." The same paragraph contrasts the replaced backend: "If the service corresponds to the main application, it will continue running, but will still be unregistered to prevent future launches at login."

After a login launch the running process is the launchd job, so flipping the toggle off in Settings → General SIGTERMs the app. Kernova installs no `SIGTERM` handler, so termination bypasses `applicationShouldTerminate`, the `quitShouldTerminateAgent` gate, and the VM save-suspend — running guests are hard-killed. Everything after `unregister()` in `LoginItemService.setEnabled(_:)` (the status read, the log, `refreshFromStatus()`) also never runs.

**2. An app update can silently stop the login launch from working.** The `SMAppService` class discussion states: "If an app updates either the plist or the executable for a LaunchAgent or LaunchDaemon, the SMAppService must be re-registered or it may not launch. It is recommended to also call unregister before re-registering if the executable has been changed." The agent's executable is the main Kernova binary, so every release changes it, and registration happens only on an explicit off→on toggle. `status` keeps reporting `.enabled`, so the switch reads on and the pane gives no signal. `SMAppService.mainApp` carried no such requirement. The recommended unregister-before-re-register is itself unsafe while behavior 1 stands.

**3. Two instances can start at login.** `AppDelegate.isAnotherInstanceRunning` runs before `NSApplication.shared`, i.e. before this process registers with Launch Services. A session-restore launch fires at the same moment as the launchd job: the job's guard cannot see a restore process that has not registered yet, and Launch Services cannot fold the restore into the job process for the same reason. Both survive, and nothing locks the VM library — `LaunchPosture.yields(toRunningInstance:)` is deliberately one-sided, so the restore process (`.manual`) never yields. "Reopen windows when logging back in" plus the toggle on is an ordinary configuration.

## Location
`LoginItemService.setEnabled(_:)`, `LoginAgentRegistration`, `AppDelegate.isAnotherInstanceRunning`, `LaunchPosture.yields(toRunningInstance:)`, `Kernova/App/app.kernova.loginlaunch.plist`.

## Considerations
All three follow from one structural fact — the registered service is the VM-hosting process — so they likely share a resolution. #796 decided the plist runs the app itself with `--login-launch`, so changing that is maintainer sign-off territory. Options raised in review, none evaluated here: a short-lived bootstrap job that hands off to the app (which would also restore Launch Services dedupe, addressing 3); a graceful `SIGTERM` path; deferring or gating `unregister()`; a library-level lock independent of launch route.

Design when picked up.

## Decision (maintainer, 2026-08-11): register the main app with `SMAppService.mainApp` — no headless login posture

Supersedes the LoginItem + bootstrap direction (its transport blocker — a sandboxed Launch Services caller cannot deliver launch arguments or environment — is documented in the comments). Rather than defending the headless posture with new entitlements and failure modes, drop it: *Open at Login* registers the main app via `SMAppService.mainApp`, and a login launch is an ordinary launch — `.regular`, library window on screen — identical to a user double-click.

All three findings resolve by SDK contract:

1. `unregisterAndReturnError` on a main-app service leaves a running app running — the toggle can never kill guests.
2. The re-register-after-update rule is scoped to LaunchAgents/LaunchDaemons; `mainApp` carries no such requirement.
3. A `mainApp` login launch is LS-mediated, so Launch Services folds it with a session-restore launch.

Scope:

- `LoginItemService` keeps the `LoginItemRegistration` protocol seam, now over `SMAppService.mainApp`.
- Removed: the agent plist and its build phase, `LoginAgentRegistration`, `LaunchPosture`, the `--login-launch` argv contract, and the login-launch duplicate-instance backstop from #796/#800 (verify via git history that no other flow depends on the instance check before deleting it).
- The Settings item now appears in System Settings' "Open at Login" section proper; update the pane's caption/affordances accordingly. The status-driven flow (including `.requiresApproval` deep-linking) stays.
- PR #803 (LoginItem helper) is superseded and closed.
- Lands in one PR together with #793's decided off-behavior; the PR closes both.

nicholas-lonsinger:
Closes #801.

## Summary

Registering the app's own executable as the login LaunchAgent (#796/#800) put three documented `SMAppService` behaviors onto the VM-hosting process:

1. `unregister` kills a running registered service, so turning *Open at Login* off SIGTERMed the app — bypassing `applicationShouldTerminate`, the `quitShouldTerminateAgent` gate, and the VM save-suspend — and killed `setEnabled(_:)` mid-call before its status read.
2. The SDK requires re-registration after a LaunchAgent's executable changes. That executable was the main binary, so every release silently broke the login launch while `status` kept reporting `.enabled`.
3. launchd `execv`s the bundle executable directly, bypassing Launch Services' dedupe, so a session-restore launch and the launchd job could both survive.

Per the maintainer's decision on #801, the registered service is now a bootstrap helper instead of the app. All three follow from the single structural fact that the registered service was the long-running process, and removing that fact resolves all three: `unregister` reaches only the helper, which is never running when the toggle flips; the re-register rule is scoped to LaunchAgents and LaunchDaemons, while a LoginItem resolves through Launch Services at login against the bundle on disk; and both routes into the app are now Launch Services-mediated, so concurrent launches fold into one instance.

## Changes

- **`KernovaLoginHelper`** — new `LSUIElement` app target embedded at `Contents/Library/LoginItems`, registered by `SMAppService.loginItem(identifier: "app.kernova.loginhelper")`. At launch it opens the enclosing `Kernova.app` with `--login-launch` in `NSWorkspace.OpenConfiguration.arguments` (`activates = false`, so the headless posture #796 established is preserved) and exits.
  - Sandboxed with `app-sandbox` **alone**, deliberately not the `app-sandbox` + `inherit` pair `KernovaRelaunchHelper` carries: `inherit` describes a process the sandboxed app `exec`'d, and this one is started by the system with no parent to inherit from. Opening an application through Launch Services is brokered and needs no entitlement.
  - It derives the host app's URL from its own bundle path rather than looking the identifier up, so it always opens the copy that contains it — a Launch Services identifier lookup can elect a different copy (what `Tools/ghosts.sh` diagnoses).
  - Runs `Tools/set-build-number.sh app`, so its `CFBundleVersion` and `MARKETING_VERSION` equal the host app's. A nested bundle whose versions disagree with its host fails App Store Connect validation, which is why the guest agent's independent-version precedent does not apply here (that ships as a DMG resource, not a nested bundle).
- **`LoginItemService`** — `LoginAgentRegistration` becomes `LoginHelperRegistration` over `SMAppService.loginItem(identifier:)`. The `LoginItemRegistration` protocol seam, `setEnabled(_:)`, and the `.requiresApproval` deep-link flow in `GeneralSettingsViewController` are unchanged.
- **Removed** `Kernova/App/app.kernova.loginlaunch.plist` and the `Embed Login LaunchAgent` shell-script build phase.
- **`LaunchPosture` and the `--login-launch` argv contract are unchanged.** `LaunchPosture.yields(toRunningInstance:)` and `AppDelegate.isAnotherInstanceRunning` are kept and demoted to a backstop in their doc comments: both run before `NSApplication.shared`, so they still cover the window in which a peer instance exists that Launch Services could not yet fold.
- **Tests** — `LoginHelperBundleTests` replaces `LoginLaunchAgentPlistTests`, asserting the built helper's path, that its shipped `CFBundleIdentifier` is the string the registration asks for, `LSUIElement`, version parity with the host, and a runnable executable. `LaunchPostureTests` gains a pin on the flag literal, which the helper necessarily duplicates.

## Deviations from the issue

The issue's Decision section fixed the design; these are gaps it left open.

- **Naming.** Target `KernovaLoginHelper`, bundle id `app.kernova.loginhelper`, following the existing `app.kernova.relaunchhelper` / `app.kernova.macosagent` shape. `SMAppService.loginItem(identifier:)` takes the helper's bundle identifier, and a test pins the shipped bundle to the registration's string.
- **The `--login-launch` literal is duplicated** into the helper. A `PBXFileSystemSynchronizedRootGroup` spans one folder and `membershipExceptions` only subtracts, so the two targets cannot share the declaration. Moving `LaunchPosture` into `KernovaKit` would drag a framework into a 60-line helper; a shared single-file root group is the move if this contract ever grows past one constant.
- **No migration or stale-registration cleanup.** The LaunchAgent has one commit in its history and there are no tags, so it never shipped; AGENTS.md reserves migration for old-shape data confirmed to exist. See the manual step below for a machine that registered it while testing #800.
- **A LaunchAgent-absence assertion was dropped from the tests.** `Contents/Library/LaunchAgents` survives in an incremental build arena as unpruned output of the removed phase, so asserting its absence fails on any machine with a pre-change build rather than signalling a defect.
- **AGENTS.md logging table** — the `app.kernova` row named the host app and KernovaKit but not embedded helpers, which `KernovaRelaunchHelper` already contradicted. Widened by one clause rather than adding a second exception.

## Test plan

- [x] `make build`, `make lint`, `make test` (full suite green; the five new tests verified to actually execute, not filter to nothing)
- [x] Helper lands at `Contents/Library/LoginItems/KernovaLoginHelper.app`, identifier `app.kernova.loginhelper`, `LSUIElement` true, `CFBundleVersion`/`CFBundleShortVersionString` equal to the host app's, signed with only `com.apple.security.app-sandbox` (plus Debug's `get-task-allow`)

Needs a real login round-trip, so unchecked for the maintainer:

- [ ] **Argument delivery** — does the launched app actually see `--login-launch` in argv when a sandboxed helper opens a sandboxed app via `OpenConfiguration.arguments`? Confirm the app logs the headless-`.accessory` line; watch with `/usr/bin/log stream --predicate 'subsystem BEGINSWITH "app.kernova"'`.
- [ ] **Finding 1** — toggle *Open at Login* off with a VM running; the app must stay alive and the switch settle to off.
- [ ] **Finding 2** — toggle on, quit, replace the app bundle with a rebuild, log out and back in; the login launch must still fire with no re-registration.
- [ ] **Finding 3** — with "Reopen windows when logging back in" enabled and a Kernova window open, log out and back in; exactly one `pgrep -x Kernova`.
- [ ] System Settings → General → Login Items & Extensions still lists Kernova under **Allow in the Background**.

One-time step on a machine that registered the old LaunchAgent while testing #800: toggle *Open at Login* off in the old build before installing this one — the registration does not self-clean when the plist disappears.

https://claude.ai/code/session_01XiN972iTjjaaMSeuhMNydK


nicholas-lonsinger:
**Converted to draft — blocked on a design decision.** See https://github.com/nicholas-lonsinger/kernova/issues/801#issuecomment-5260563193 for the full analysis.

Short version: `NSWorkspace.h` documents `OpenConfiguration.arguments` as **ignored when the calling process is sandboxed**, and the helper must be sandboxed. I tested the neighbouring `environment` property as a substitute and it is dropped identically. The same call from an unsandboxed process delivers both, so the sandbox is the cause, not the call.

That means the `--login-launch` transport in this diff never fires on the path that matters: the app reads `.manual` and shows the library window at login, the exact behavior #796 removed. The argv channel worked under #800 only because launchd `execv`'d the app directly rather than going through a sandboxed Launch Services caller.

Everything else in the diff stands and is verified — the embedded `KernovaLoginHelper` target, `SMAppService.loginItem` registration, removal of the agent plist and its build phase, docs, and tests (build, lint, and full suite green; helper verified at the correct path, identifier, `LSUIElement`, version parity, sandboxed and signed). The branch is ready to receive whichever transport you pick.

nicholas-lonsinger:
Superseded by a direction change on #801: the headless login posture is dropped entirely — *Open at Login* now registers the main app via `SMAppService.mainApp` (decision updated on the issue). The helper/bootstrap design this PR built is no longer needed. The transport findings documented on #801 remain the record of why.

PR #804, opened 2026-08-12
fix: Register Open at Login as the main app, and quit when keep-off

### LINKED ISSUE #793: Replace the General pane's two residency toggles with a single three-rung setting
## Found during
2026-08-10 session: the maintainer toggled *Keep Running in Menu Bar* off expecting the status icon to go away; it stayed.

## Description
`keepInMenuBarOnQuit` currently governs only what a GUI-origin quit does (`quitShouldTerminateAgent`, `appMenuQuitItems`). The status item is created unconditionally in `startResidentApp()` and nothing observes the preference, so with the toggle off the app still shows the menu-bar icon and still goes headless-resident (`.accessory`) when the last window closes.

Decided behavior (maintainer, 2026-08-10): **off means behave like a normal app** —

- No status item while the toggle is off (and it appears/disappears live when the toggle flips).
- When the last window closes, the app quits (save-suspending running VMs, as a full quit does today) instead of demoting to `.accessory` — with no Dock icon and no status item it would otherwise be unreachable.
- Shortcuts adjust to match ("adjust any shortcuts as well"): the app menu already collapses to a single ⌘Q *Quit Kernova* when the toggle is off; audit for other keep-mode-only affordances that reference the status item (e.g. surfaces that direct the user to the menu-bar icon).

Open question: what should flipping the toggle off do when the app is already headless-resident at that moment (Settings is a window, so in practice at least one window is open — but the transition ordering needs a decision).

## Location
`AppDelegate.startResidentApp()` (status-item creation), `quitShouldTerminateAgent`, `appMenuQuitItems`, the window-close reconcile path (`hasVisibleUserWindow` / `syncAgentActivationPolicy`), `GeneralSettingsViewController.keepInMenuBarToggled`.

## Suggested fix
Observe the preference; tear down / create the status item on change; in off mode the last-window-close reconcile terminates through the existing quit gate instead of demoting. Design when picked up.

## Decision (maintainer, 2026-08-11): keep two independent toggles

#801's final direction registers the main app via `SMAppService.mainApp`, so a login launch opens the app normally, window on screen. That dissolves the premise of the residency-ladder idea previously recorded here (a login launch no longer starts with zero windows), so *Open at Login* + *Keep Running in Menu Bar* off is coherent again: opens at login like a normal app, quits when its last window closes. The two switches (as renamed in #802) stay independent; no ladder.

What remains is the decided off-behavior above, unchanged: no status item while the toggle is off, created/torn down live on the flip; last-window-close terminates through the existing quit gate (save-suspending running VMs) instead of demoting to `.accessory`; audit keep-mode-only affordances that reference the status item. The open transition-ordering question resolves itself: Settings is a window, so the flip always happens with a window open — tear down the status item immediately, and the quit-instead-of-demote rule applies at the next last-window-close.

Lands in one PR together with #801; the PR closes both.

### LINKED ISSUE #801: Registering the app itself as the login LaunchAgent collides with its own process lifecycle

nicholas-lonsinger:
Closes #801, closes #793.

Both issues' Decision sections land together because they are coupled: #801 drops the headless login posture, which is what dissolved the residency-ladder premise in #793 and left two independent toggles coherent.

## #801 — the login registration

Registering the app's own executable as the login LaunchAgent (#796/#800) put three documented `SMAppService` behaviors onto the VM-hosting process:

1. `unregister` kills a running registered service, so turning the toggle off SIGTERMed the app — bypassing `applicationShouldTerminate`, the `quitShouldTerminateAgent` gate, and the VM save-suspend.
2. The SDK requires re-registration after a LaunchAgent's executable changes; that executable was the main binary, so every release silently broke the login launch while `status` still read `.enabled`.
3. launchd `execv`s the bundle executable directly, bypassing Launch Services' dedupe, so a session-restore launch and the launchd job could both survive.

`SMAppService.mainApp` resolves all three by contract: `unregisterAndReturnError` leaves a running main app running, the re-register rule is scoped to LaunchAgents and LaunchDaemons, and a `mainApp` login launch is Launch Services-mediated and therefore folded with a session restore.

The cost, and the decided trade: **there is no headless login posture any more.** A login launch is `.regular` with the library window on screen, identical to a double-click. That is what lets the entire `--login-launch` mechanism go — the plist, its build phase, `LaunchPosture`, and the argv handling in `main()`.

The previous direction (a LoginItem + bootstrap helper, PR #803) is abandoned; its transport blocker — a sandboxed Launch Services caller cannot deliver `arguments` **or** `environment`, verified with a controlled A/B — is recorded on #801.

## #793 — what "off" means

*Continue running in Status Bar* governed only `quitShouldTerminateAgent`. The status item was created unconditionally in `startResidentApp()` and the last-window-close reconcile always demoted to `.accessory`, so turning the toggle off left the icon on screen and the app resident — the behavior that prompted the issue.

Off now means the app behaves like a normal Mac app: no status item, and the last window close quits.

## Changes

- **`LoginItemService`** — `LoginAgentRegistration` → `MainAppRegistration` over `SMAppService.mainApp`. The `LoginItemRegistration` protocol seam, `setEnabled(_:)`, and the `.requiresApproval` deep-link flow are untouched.
- **Deleted** — `Kernova/App/LaunchPosture.swift`, `app.kernova.loginlaunch.plist` and its `Embed Login LaunchAgent` build phase, the posture branch in `main()` and `startResidentApp()`, and `AppDelegate.isAnotherInstanceRunning`. Checked before deleting the instance guard, as #801's Decision asks: `git log -S` shows both it and `LaunchPosture.yields` first appear in `685cdc6` (#800) and the tree has no other caller, so they leave with the posture they guarded.
- **`AppDelegate.residencyOutcome(hasVisibleUserWindow:keepInMenuBar:hasPreparingInstance:)`** — a `nonisolated static` decision function returning `.showDockIcon` / `.goHeadless` / `.quit`, following the existing testable-helper shape of `appMenuQuitItems` and `statusItemOpenTarget`. `syncAgentActivationPolicy()` switches on it; the `.quit` branch calls `NSApp.terminate(nil)`, so VMs save-suspend through the existing gate rather than a second save path.
- **`syncStatusItem()` + `observeResidencyPreference()`** — the status item is created and torn down from an observation of the toggle. New `HostAgentStatusItemController.tearDown()` cancels the three observation loops, dismisses any transient popover (it owns a live auto-dismiss task), and calls `NSStatusBar.removeStatusItem` — the controller had no teardown at all before.
- **`VMLibraryViewModel.keepInMenuBarOnQuit`** — an observable mirror of the preference, and its single write path. `AppPreferences` is a struct over `UserDefaults`, which `withObservationTracking` never wakes; this follows the `agentInstallPromptDisabled` precedent, whose doc comment already records that reasoning. `AppDelegate` observes rather than the Settings pane calling into the delegate — a view→delegate back-edge exists nowhere in this codebase.
- **Settings copy** — the login row is *Open at Login* again (see below); the residency caption gains one sentence naming the off-behavior.

## Calls the Decisions did not cover

- **The toggle names are stale in both Decision sections.** They say "Open at Login" and "Keep Running in Menu Bar", but `3d75aed` (#802, merged the same morning) renamed those *to* "Launch into Background at Login" and "Continue running in Status Bar". Split the difference on the merits: the login toggle goes back to **Open at Login**, because "Launch into Background at Login" becomes factually false once the headless posture is deleted and #801's Decision grounds the new behavior in System Settings' own "Open at Login" section. The residency toggle **keeps its #802 name** — #793's Decision says the toggles stay "as renamed in #802", and nothing here falsifies that label.
- **An in-flight import vetoes the quit-on-close.** Neither Decision covers it, and without a guard it is data loss: the quit path runs `cancelAndCleanupPreparingInstances()`, which trashes partial bundles, so closing a window during an import would silently destroy it. A *running* VM deliberately does not veto — save-suspending it is the decided behavior. When the import finishes the observation re-runs the reconcile, so the quit is deferred rather than cancelled.
- **Re-entrancy latch.** `applicationShouldTerminate` returns `.terminateLater` while VMs save; a window closing in that window would otherwise request a second termination against an outstanding reply. `isQuittingOnLastWindowClose` latches and is never reset, mirroring the existing quit flags.
- **`applicationShouldTerminateAfterLastWindowClosed` still returns `false`.** The reconcile keys on `hasVisibleUserWindow`, which counts miniaturized windows and untracked panels (the About panel) that AppKit's own last-window rule does not; letting AppKit terminate too would double-fire on a different predicate. Comment updated to say so.
- **Closing the Settings window can now quit the app** when it is the last window and the toggle is off. An unavoidable consequence of "behave like a normal app", and the flip itself is safe because Settings is a window — the reconcile sees it and returns `.showDockIcon`, which is how #793's open transition-ordering question resolves.
- **Clipboard notices while the status item is absent** — `ClipboardIssueCenter.pendingNotice` has no consumer with the toggle off. It is a single-slot latch, not a growing queue, and the clipboard window's own banner still reads it. No change; noted so a reviewer does not re-derive it as a leak.
- **Audited the keep-mode-only affordances** #793 asks about. Three surfaces reference the status item and all already self-gate: the soft-quit reminder is reachable only through the keep-on branch, the Reminders pane conditions its own copy, and the app menu already collapses to a single ⌘Q. No changes needed.

## Test plan

- [x] `make build`, `make lint`, `make test` — full suite green; the new suites verified to actually execute rather than filter to nothing
- [x] **Toggle on**, last window closed on the real app: logs `Activation policy 0 → 1 (hasVisibleWindow=false)`, process stays alive
- [x] **Toggle off** (injected through `NSUserDefaults`' argument domain, since the preference lives in a TCC-protected container), last window closed: logs `Last window closed with the app set to quit — terminating`, process exits
- [x] Built bundle no longer ships `Contents/Library/LaunchAgents`

Needs a real login round-trip or the Settings UI, so unchecked for you:

- [ ] Toggle *Open at Login* on → Kernova appears in System Settings → General → Login Items under **Open at Login** (not "Allow in the Background")
- [ ] Log out and back in → the app opens with its library window, exactly one instance (`pgrep -x Kernova`)
- [ ] Toggle *Open at Login* off with the app running and a VM up → nothing is killed, the System Settings entry disappears
- [ ] Flip *Continue running in Status Bar* off in Settings → the status icon disappears immediately; flip it on → it comes back
- [ ] With the toggle off and a VM running, close the last window → the app quits and the VM save-suspends
- [ ] All four combinations of the two toggles behave coherently

One-time step on a machine that registered the old LaunchAgent while testing #800: toggle the login item off in the old build before installing this one — the registration does not self-clean when the plist disappears.

### Maintenance Notes

- ✅ **ARCHITECTURE.md** — updated. The `HostAgentStatusItemController` bullet now records that the status item's lifetime is conditional on the toggle (a changed component lifetime, which is what clears the protocol's bar); the `LoginItemService` mention and the ServiceManagement dependency row move to `SMAppService.mainApp`. No target added or removed.
- ✅ **Tests** — `AppDelegateResidencyOutcomeTests` covers the new decision function including the import veto; `VMLibraryViewModelTests` gains seeding, write-through, and an event-driven observation-wake case for the mirror; `LoginItemServiceTests` keeps its `FakeRegistration` seam and gains one case pinning the shipping registration to `mainApp`. `LaunchPostureTests` and `LoginLaunchAgentPlistTests` are deleted with the types they covered. `HostAgentStatusItemController.tearDown()` touches `NSStatusBar` and is not unit-testable here — consistent with that controller having no tests today; it is in the manual list above.
- ✅ **AGENTS.md** — no rule stated there changed.
- ✅ **Other docs** — SANDBOX.md's launch-model paragraph and README's feature bullet rewritten; no `MARKETING_VERSION` bump, as the guest agent is untouched.

https://claude.ai/code/session_01XiN972iTjjaaMSeuhMNydK


nicholas-lonsinger:
## Review triage

Three lanes at high effort: the built-in `/code-review`, a Codex branch review, and a Codex adversarial review. The `SMAppService.mainApp` half came back clean in all three — no dangling references, the pbxproj phase removal is complete, and the mirror matches the `agentInstallPromptDisabled` precedent. Everything below is the new quit-on-last-window-close reconcile, fixed in 143d9cb.

**Fixed now**

| Finding | Lanes | Why it cleared the bar |
|---|---|---|
| A hidden app (⌘H) reads as having no windows, so a background close quits with the library still open | built-in + Codex review | `hasVisibleUserWindow` tests `isVisible`, which is false for every window of a hidden app — its own doc comment names the scenario ("a VM shutting down empties the last display window mid-hide"). Under `main` the outcome was a harmless demote; this branch had turned it into a termination. Reachable by ⌘H plus a guest shutting itself down. |
| A VM mid-save/restore/start/install is neither held back nor saved | built-in + adversarial | `applicationShouldTerminate` filters to `.running`/`.paused`, so a `.saving` VM falls through both the hold and the save loop and the process exits mid-`saveMachineState`. ⌘W during a suspend truncated the save file. |
| A held quit demoted to `.accessory` | built-in | Produced exactly the unreachable state `residencyOutcome`'s own doc rules out: with the toggle off there is no status item, so an in-flight import lost both its progress readout and its cancel affordance. Now `.showDockIcon`. |
| Pop In races the reconcile into a quit | adversarial | The global close observer schedules a reconcile that runs *before* the pop-in handler's restore, sees zero windows, and terminates. The pop-in handler's existing comment already documented that ordering as unspecified. The global observer now skips a display window closing for `.popIn`, which also removes the redundant reconcile behind the Dock-icon flicker that comment describes. |
| Two sources of truth for the residency preference | built-in | `quitShouldTerminateAgent` and `rebuildAppMenuQuitItems` read `AppPreferences` directly while the new code read the mirror; they agreed only because the Settings switch is the sole writer. Both now read the mirror. |

**Deviation from a recommendation.** Two lanes proposed `VMInstance.isKeepingAppAlive` as the hold predicate. That would also veto a settled `.running` VM, contradicting the decided behavior — save-suspending running VMs on the last-window close is the point of #793. The hold is `isPreparing || status.isTransitioning` instead, which is `isKeepingAppAlive` minus the settled states the termination gate already handles correctly. `VMStatus.isTransitioning` already existed (untested, used by `canRename`); it gained the doc explaining the exclusion and became exhaustive so a new state has to choose a side.

**Dismissed**

- `pasteProgressPresenter` can leave one pending `performOnMainRunLoop` calling `performClick` on a removed status item. Inert, and below the severity bar.
- `ClipboardIssueCenter.pendingNotice` going unconsumed while the toggle is off — a single-slot latch, not a growing queue, and the clipboard window's own banner still reads it. Already noted in the PR body.
- The pre-existing termination gap on the explicit-quit path (⌥⌘Q during a save has the same shape). Out of scope here; this PR only closes the new route to it. Worth its own issue if you want the gate itself made lifecycle-aware.

No new issues filed — everything actionable was in scope and is fixed on the branch.

Issue #815, opened 2026-08-12
feat: Host-only and VM-to-VM networking


nicholas-lonsinger:
## Summary

Add a host-only networking mode: guests reachable from the host (and optionally each
other) on stable private addressing, with no LAN exposure and no internet route. Revives
#279, whose closing comment requested this issue when the entitlement landed.

## Context (carried forward from #279's spike — verified then, re-verify on pickup)

- `vmnet_network_create` succeeds host-mode/shared-mode unsandboxed and fails under the
  sandbox without the entitlement — an XPC sandbox denial (com.apple.NetworkSharing)
  surfacing as VMNET_MEM_FAILURE. Host-only does NOT dodge the entitlement.
- Under today's NAT attachment, host↔guest already works; guest↔guest is blocked by
  vmnet's default isolation. An airgapped VM (no network device) works today.
- macOS 26 provides the sanctioned path: vmnet_network_configuration_create(host mode) +
  VZVmnetNetworkDeviceAttachment. The app's floor is macOS 26 — no gating.
- Isolation is expressed by network membership, not a flag: VMs on the same custom
  network reach each other; separate networks are mutually isolated. The per-interface
  vmnet_enable_isolation_key is not reachable through the VZ attachment path — strict
  host-only means one network per VM, VM↔VM means a shared one.

## Scope

- An app-managed host-mode vmnet network; "Host Only" joins the per-VM mode control.
- Persistence: vmnet networks do not survive app relaunch — serialize
  (vmnet_network_copy_serialization) and rebuild at launch so addressing stays stable.
- Decide and document (in #814's terms) the default topology: a shared host-only network
  (host + VM↔VM) vs one network per VM (strict host-only), and whether both are offered.

## Considerations

- Network configuration is immutable while active (Apple DTS) — edits require the
  network's VMs stopped, matching the existing lock-while-running settings pattern.
- This sub-issue carries the custom-network backbone (creation, serialization, attachment
  plumbing) that #816 and #817 extend.

## Related

Parent: #819. Depends on #811, #814. Supersedes #279.

- UI direction: docs/research/2026-08-12-network-settings-ui.md (this issue adds the
  Host Only picker entry). Deviations need a stated reason in the PR.
- Remaining implementation design intentionally omitted — design when picked up.

Issue #85, opened 2026-03-25
feat: Add container support alongside virtual machines


nicholas-lonsinger:
## Summary

Add first-class container support to Kernova, allowing users to manage lightweight Linux containers alongside full virtual machines. The sidebar should present VMs and containers as separate collapsible groups, each with its own header.

## Motivation

Virtual machines provide strong isolation but carry overhead (full OS, disk images, boot time). Containers offer a lightweight alternative for many Linux workloads — faster startup, lower resource usage, and simpler lifecycle. Supporting both in a single app lets users pick the right tool for each task without leaving Kernova.

## Sidebar Design

The sidebar currently shows a flat list under a single "Virtual Machines" section. This feature introduces **two collapsible groups** with disclosure indicators:

```
▾ Virtual Machines
    macOS Sequoia
    Ubuntu Server
▾ Containers
    nginx-dev
    postgres-16
```

- Each group header is a collapsible `DisclosureGroup` (or equivalent) with a title and item count badge
- Collapse state should persist across launches (UserDefaults)
- Empty groups show a subtle "No items" placeholder or hide entirely (TBD)
- Drag-and-drop reordering works independently within each group
- Selection spans both groups (selecting a container deselects any VM and vice versa)

## Container Model

Introduce a container representation parallel to `VMInstance` / `VMConfiguration`:

| Concept | VM equivalent | Container equivalent |
|---------|--------------|---------------------|
| Runtime wrapper | `VMInstance` | `ContainerInstance` |
| Persistent config | `VMConfiguration` | `ContainerConfiguration` |
| Status enum | `VMStatus` | `ContainerStatus` (created, starting, running, stopped, error) |
| Bundle on disk | `.kernova` package | `.kernovacontainer` package (or subdirectory) |
| Sidebar row | `VMRowView` | `ContainerRowView` |

### ContainerConfiguration (persisted as `config.json`)

Key fields to consider:
- `id: UUID`, `name: String`, `createdAt: Date`
- `image: String` — OCI image reference (e.g., `docker.io/library/ubuntu:24.04`)
- `cpuCount: Int`, `memorySizeInMB: Int` — resource limits
- `portMappings: [PortMapping]` — host↔container port forwards
- `environmentVariables: [String: String]`
- `volumes: [VolumeMount]` — host directory bind mounts
- `entrypoint: String?`, `command: [String]?`
- `networkMode: ContainerNetworkMode` (bridge, host, none)

### Options to evaluate

[...]

PR #832, opened 2026-08-14
feat: Port forwarding for shared networking

### LINKED ISSUE #817: feat: Port forwarding for shared networking
## Summary

Per-VM port-forwarding rules for shared networking: reach a guest service from outside
without bridged mode. The Network section's info text names this exact gap today.

## Context

- macOS 26 provides config-time rules on a custom vmnet network:
  vmnet_network_configuration_add_port_forwarding_rule (TCP and UDP, IPv4 and IPv6).
  Two argument-order traps: internal/external are transposed relative to the older
  interface-level API, and Apple's own @param doc-comment lists a different order than
  the actual signature — the signature wins.
- Known platform limitation on macOS 26: a forwarded port is reachable from other
  machines but NOT from the host itself via localhost (Apple DTS: "a known limitation of
  vmnet", FB7731708; fixed in macOS 27). The rules UI must disclose this inline on
  macOS 26 rather than letting users debug it.
- Runtime rule mutation may be unavailable through the VZ attachment path: the mutation
  APIs need an interface_ref the attachment never exposes, and config-time rules cannot
  be removed or queried until the network starts. Inference from headers, not observed.
- Rule-table conventions across Fusion/UTM/Lima: symmetric Host/Guest columns, host side
  defaulting to the guest side, entry-time validation (VirtualBox accepts a sub-1024 host
  port at edit time and the VM then refuses to start — validate at entry instead).

## Scope

- A rules table on VMs using shared networking (Protocol / Host Port / Guest Port),
  editable while the VM is stopped, validated at entry.
- Conservative binding defaults per #814.

## Open questions

- Spike first: are rules runtime-mutable at all via VZVmnetNetworkDeviceAttachment, or
  strictly config-time? The answer decides whether the table is live or stopped-only.

## Related

Parent: #819. Depends on #815. #85's container port mappings should reuse this model.

- UI direction: docs/research/2026-08-12-network-settings-ui.md (this issue adds the
  in-card Port Forwarding rows and the Add Rule sheet). Deviations need a stated reason
  in the PR.
- Remaining implementation design intentionally omitted — design when picked up.

nicholas-lonsinger:
Closes #817

## Summary

Per-VM port-forwarding rules for Shared Network VMs: a guest service becomes reachable from other machines on the host's network without bridged mode. Rules are installed on the app-managed vmnet shared network at creation, forwarding to each VM's reserved (deterministic) address — the same machinery that carries DHCP reservations (#815/#816).

The spike the issue asked for is recorded in [this comment](https://github.com/nicholas-lonsinger/kernova/issues/817#issuecomment-5288071006): rules are **strictly config-time** through the VZ attachment path — every runtime mutation API needs an `interface_ref` the attachment never exposes. That decided the design:

- **Rules table is stopped-only** (the section's existing lock behavior).
- The service tracks *desired* vs *installed* rules and exposes `portForwardingRulesArePending(for:)`; the library recreates the shared network whenever rules pend and no live session could be attached to it (rule edits, session teardown). The recreate rides the existing pinned-addressing machinery, so no guest's address moves — and an edited rule takes effect at the next VM start without an app relaunch, even though the IP-address row materializes the network early.
- Each VM's `config.json` is the single source of truth (`portForwardingRules` on `VMConfiguration`). The service's per-MAC declarations are in-memory only, re-declared at library load — deliberately *not* persisted to `networks.json`, which would be a second source of truth (reservation slots are persisted there only because slot *order* is genuinely owned state).
- One forwarding model (`PortForwardingRule` / `PortForwardingTransport` / `PortForwardingHostClaim`), per docs/NETWORKING.md principle 4 — #85's container mappings should reuse it.
- A host port is claimed network-wide, so the Add Rule sheet validates collisions across **all** Shared Network VMs' rules, not just the edited VM's; the service additionally dedupes deterministically (reservation-slot order, first wins, logged) so a duplicate can never reach vmnet and fail network creation for every VM on it.
- The vmnet call follows the SDK **signature** — internal (guest) port before external (host) port — against the header's contradictory `@param` order; commented at the call site.

UI per docs/research/2026-08-12-network-settings-ui.md decision 6: rule rows inside the Network card beneath MAC (`TCP  Host 8080 → Guest 80`, borderless remove), a trailing `Add Rule…` row opening a compact sheet (Protocol, Host Port, Guest Port; Add disabled while a port is empty/zero/out-of-range or the host port is taken; sub-1024 inline note; host port mirrors the guest port until edited). Rows show only for entitled Shared Network VMs. The macOS 26 loopback limitation is disclosed in the section's info popover.

## Deviations from the design doc (stated per its own rule)

- **Rule rows render as the mock does** — indented lines under a "Port Forwarding" title with no hairline between rules — rather than the prose's "one hairline row per rule". The mock and prose disagree; the mock's rendering keeps the card's major-row hairlines meaningful.
- **Rows additionally require a MAC address** (matching the IP row's gating): without a MAC there is no reservation to forward to, so the control would be visible but dead.
- **The popover's entitled-build clause was reworded, not just appended**: it asserted "there is no port forwarding from host to guest", which this feature makes false.
- **NETWORKING.md principle 3's "narrowest useful listener" cannot be implemented**: `vmnet_network_configuration_add_port_forwarding_rule` has no bind-address/listener parameter — the listener scope is fixed by vmnet (reachable from the external network; on macOS 26, *not* from the host's own localhost). The popover discloses the actual reach instead.

## Changes

- `Kernova/Models/PortForwardingRule.swift` (new) — the forwarding model
- `Kernova/Models/VMConfiguration.swift` — `portForwardingRules` (`decodeIfPresent ?? []`)
- `Kernova/Models/VMStatus.swift` — `isResting`; `Kernova/Models/VMInstance.swift` — `onSessionEnded`, fired when a VM settles into a resting status
- `Kernova/Services/VmnetNetworkService.swift` — rule install at `createNetwork`, desired/installed rule state, `setPortForwardingRules`, `portForwardingRulesArePending`; slot resolution factored into one shared path
- `Kernova/ViewModels/VMLibraryViewModel.swift` — `syncPortForwardingRules` beside the reservation sync (update funnel + instance construction), eviction cleanup, `rebuildSharedNetworkIfIdle`
- `Kernova/Views/Detail/VMSettingsViewController.swift` — Port Forwarding rows, popover copy; `PortForwardingRuleSheetContentViewController.swift` (new) — the Add Rule sheet
- Tests: `VmnetNetworkServiceTests`, `VMConfigurationTests`, `VMLibraryViewModelTests`, `VMSettingsViewControllerTests`, new `PortForwardingRuleSheetContentViewControllerTests`, mock extensions
- `docs/ARCHITECTURE.md` — one clause on the `VmnetNetworkService` paragraph

## Test plan

- [x] Full suite green (2593 test cases, 0 failures)
- [x] Service: install with slot-derived internal address, cross-VM dedupe, MAC-without-slot drop, pending-flag semantics (installable vs installed, so an uninstallable rule can't wedge it), invalidate clears installed rules
- [x] Model: Codable round-trip, absent-key default, clone carries rules
- [x] View model: sync on update/construction, clear on mode switch and evict, idle rebuild fires only when pending and no shared VM is live
- [x] UI: row rendering/gating (mode, entitlement, MAC), remove writes config, read-only lock; sheet validation matrix incl. collision note and guest→host prefill

https://claude.ai/code/session_01JVhaNysNERLWKNJRE9ECkR

### DIFF
--- a/Kernova/Models/PortForwardingRule.swift
+++ b/Kernova/Models/PortForwardingRule.swift
@@ -1,0 +1,41 @@
+import Foundation
+
+/// The transport a forwarding rule covers.
+enum PortForwardingTransport: String, Codable, Sendable, CaseIterable {
+    case tcp
+    case udp
+
+    /// How the transport reads in the UI.
+    var displayName: String { rawValue.uppercased() }
+}
+
+/// One host→guest port mapping on an app-managed network.
+///
+/// The app's single forwarding model (docs/NETWORKING.md): every consumer of
+/// port mappings uses this rule and the enforcement path behind it.
+struct PortForwardingRule: Codable, Sendable, Equatable, Hashable {
+    /// The transport forwarded.
+    var transport: PortForwardingTransport
+    /// The port traffic arrives on, on the host side (vmnet's *external* port).
+    var hostPort: UInt16
+    /// The port traffic is delivered to inside the guest (vmnet's *internal*
+    /// port).
+    var guestPort: UInt16
+
+    /// The ports a rule may carry — port 0 addresses no service.
+    static let portRange: ClosedRange<Int> = 1...65535
+
+    /// What makes this rule collide with another.
+    var hostClaim: PortForwardingHostClaim {
+        PortForwardingHostClaim(transport: transport, hostPort: hostPort)
+    }
+}
+
+/// A claim on one host-side port.
+///
+/// A network carries one rule per (transport, host port): the host port is
+/// claimed network-wide, across every VM joined to it.
+struct PortForwardingHostClaim: Hashable, Sendable {
+    var transport: PortForwardingTransport
+    var hostPort: UInt16
+}
--- a/Kernova/Models/VMConfiguration.swift
+++ b/Kernova/Models/VMConfiguration.swift
@@ -450,6 +450,10 @@ struct VMConfiguration: Codable, Sendable, Equatable {
             SharedDirectory(id: UUID(), path: dir.path, readOnly: dir.readOnly, bookmark: dir.bookmark)
         }
 
+        // A host port is claimed once network-wide, so carried-over rules would
+        // collide with the source's the moment both VMs share a network.
+        clone.portForwardingRules = []
+
         // The clone copies the source bundle's post-install artifacts, so
         // preserving either install context would falsely mark it as awaiting
         // an initial boot.
--- a/Kernova/Models/VMConfiguration.swift
+++ b/Kernova/Models/VMConfiguration.swift
@@ -78,6 +78,13 @@ struct VMConfiguration: Codable, Sendable, Equatable {
 
     var macAddress: String?
 
+    /// Host→guest port mappings for a Shared Network VM, installed on the
+    /// app-managed network when it is created.
+    ///
+    /// Ignored in every other mode: Bridged guests are reached at their own
+    /// address, Host Only ones only from this Mac.
+    var portForwardingRules: [PortForwardingRule]
+
     // MARK: - Clipboard Sharing
 
     /// When `true`, a SPICE agent console port is configured to enable clipboard
@@ -239,6 +246,7 @@ struct VMConfiguration: Codable, Sendable, Equatable {
         networkMode: VMNetworkMode = .shared,
         bridgedInterfaceIdentifier: String? = nil,
         macAddress: String? = nil,
+        portForwardingRules: [PortForwardingRule] = [],
         clipboardSharingEnabled: Bool = false,
         clipboardPassthroughEnabled: Bool = false,
         serialSocketRelayEnabled: Bool = false,
@@ -283,6 +291,7 @@ struct VMConfiguration: Codable, Sendable, Equatable {
         self.networkMode = networkMode
         self.bridgedInterfaceIdentifier = bridgedInterfaceIdentifier
         self.macAddress = macAddress
+        self.portForwardingRules = portForwardingRules
         self.clipboardSharingEnabled = clipboardSharingEnabled
         self.clipboardPassthroughEnabled = clipboardPassthroughEnabled
         self.serialSocketRelayEnabled = serialSocketRelayEnabled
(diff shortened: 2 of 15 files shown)


nicholas-lonsinger:
Three review passes ran on this branch — `/code-review high`, Codex review, and Codex adversarial review — and converged on the same defects. Triage per AGENTS.md, all addressed in e4297ae:

**Fix now (all confirmed in code before fixing):**

1. **Status was an invalid liveness proxy** (all three reviews). A save-suspended VM rests at `.paused` with `virtualMachine == nil` — the standard quit path — so the status-based idle gate deferred pending rule edits until relaunch, and `.error` can hold a live VZ VM (pause failure), risking an invalidate under a live attachment. Fixed: the teardown hook fires from `tearDownSession()` (the single release point, covering save completion), and the gate reads the session's *actual* attachment via `kind(ofNetwork:)` — which also fixes the live-switch-into-Shared case Codex found (config already Shared, attachment still on the old network), plus a re-check after `applyLivePolicy` for a live switch out.
2. **Host-port claims scoped too narrowly** (Codex + adversarial). Rules persist in config across mode switches and reactivate, so a dormant Host Only VM's rule could be double-claimed and then silently lose on return to Shared. Fixed: the Add Rule sheet counts every VM's rules regardless of mode.
3. **Clone duplicated claims** (all three). `clonedForNewInstance` regenerates every collide-capable identity but carried rules through — an instant, silently-dead duplicate. Fixed: the clone starts with no rules, same category as the other regenerated fields.
4. **Rule declared mid-materialization was lost** (adversarial, medium). A `setPortForwardingRules` landing during the vmnet round-trip saw no handle (not pending) and nothing rechecked after publish. Fixed: `network(for:)` re-checks declarations under the state lock before publishing and recreates the never-published network (bounded at 3 attempts, then publishes with a warning and leaves the pending flag to drive a later rebuild).

**Dismissed:** surfacing service-level duplicate drops in the UI (adversarial's sub-recommendation) — after fixes 2 and 3 the remaining route to a duplicate is importing a foreign bundle, a degenerate input the deterministic, logged service dedupe already backstops.

Each fix ships with a test that fails against the pre-fix code. Suite: 2598 test cases green, lint clean.

Issue #835, opened 2026-08-14
Two VMs can hold the same MAC address, colliding on their reservation and forwarding keys


nicholas-lonsinger:
## Found during

Review of PR #833, which makes the MAC address editable and so makes this reachable by typing.

## Description

Nothing checks a MAC against the rest of the library. `VmnetNetworkService` keys both DHCP reservations and desired forwarding rules on the lowercased MAC alone (`reservedMACs`, `desiredForwardingRules[kind][mac]`), so two VMs configured with the same address share one key:

- They share a single reservation slot, so both guests are promised the same address.
- Whichever VM writes last owns the `desiredForwardingRules` entry: VM A's declaration replaces VM B's rules, and later moving A off that MAC withdraws them through the same shared key, leaving B with no forwarding until something re-declares it.

Duplicate layer-2 identities on one segment are broken regardless of what the app does, so the question is whether the app should refuse the configuration outright.

## Location

`VMSettingsViewController.applyMACAddressFieldEdit()`, `VMLibraryViewModel.updateConfiguration(of:mutate:)`, `VmnetNetworkService.setPortForwardingRules(_:for:kind:)`, `VmnetNetworkService.reserveAddressIfNeeded(for:kind:)`

## Considerations

- Enforcing uniqueness centrally in `VMLibraryViewModel` covers the field, Generate, clone, and any future writer; a check in the field delegate alone does not.
- The rejection needs a surface. The MAC field's only current rejection signal is the value snapping back, which says nothing about *why* — adequate for a malformed address, whose tooltip names the format, but opaque for a duplicate.
- Carrying an address across from a VM being replaced is a legitimate intent, so whatever refuses a duplicate should leave that reachable (clearing or deleting the other VM first).

Issue #834, opened 2026-08-14
A changed or new MAC's DHCP reservation never activates on an already-materialized network


nicholas-lonsinger:
## Found during

Review of PR #833 (editable MAC address), raised independently by three reviewers.

## Description

`VmnetNetworkService.reservedAddress(for:kind:)` returns `nil` while a network is materialized and the derived slot address does not match what that materialization installed — reservations are fixed at creation, so a slot assigned afterwards is pending until the next materialization. Nothing marks a *reservation* change as requiring a recreate: `rebuildSharedNetworkIfIdle()` consults only `portForwardingRulesArePending(for:)`.

Observed consequences, on an entitled build with the network already materialized (which viewing any Shared VM's Network section causes, since the IP Address row materializes to fill itself):

- A VM whose MAC is added or changed shows `—` in the IP Address row for the rest of the app session, and its guest gets a dynamic lease instead of the deterministic reservation. Only an app restart, or an unrelated invalidation, recovers.
- This is not specific to editing: a newly created or cloned VM takes its slot the same way and lands in the same pending state. PR #833 adds a third way to reach it, and incidentally rescues the subset of cases where the VM also carries forwarding rules (their pending state triggers the existing recreate).
- Host Only has no idle-recreate path at all; `rebuildSharedNetworkIfIdle` is Shared-only.

Separately, `reservedMACs` is append-only and persisted, so a retired MAC keeps its slot forever. Releasing it is not simply a removal — slot order determines every later address, so a naive removal would shift addresses the user relies on.

## Location

`VmnetNetworkService.reserveAddressIfNeeded(for:kind:)`, `VmnetNetworkService.reservedAddress(for:kind:)`, `VmnetNetworkService.portForwardingRulesArePending(for:)`, `VMLibraryViewModel.syncAddressReservation(for:)`, `VMLibraryViewModel.rebuildSharedNetworkIfIdle()`

## Suggested fix

Expose reservation-pending state alongside the existing forwarding-pending state and generalize the idle recreate to every `VmnetNetworkKind`, so a network whose reservations changed rematerializes once no VM holds an attachment on it. Decide separately whether a retired MAC's slot can be reused without moving any other VM's address.

PR #836, opened 2026-08-14
fix: Reclaim DHCP reservation slots when a VM or its MAC goes away

### LINKED ISSUE #830: Reclaim DHCP reservation slots when a VM or its MAC goes away
## Found during

PR #829 review (`/code-review high`) of the DHCP reservation machinery behind the IP Address row.

## Description

`VmnetNetworkService`'s per-network `reservedMACs` list is append-only: `reserveAddressIfNeeded` is its sole mutator, and nothing releases a slot when a VM is deleted, when a clone's fresh MAC replaces interest in an old one, or when a MAC is regenerated (#818 will add such a flow). Slot position is the address (`host 2 + index`), so compaction is impossible in the current shape — dead slots accumulate for the life of `networks.json`, and each one is an address the network's DHCP can no longer hand out dynamically. On the /24 vmnet reserves, capacity is 253 slots; past it, new VMs' IP Address rows show the placeholder permanently. Reaching that takes years of clone/delete churn, so this is debt, not a defect.

## Location

`VmnetNetworkService.reserveAddressIfNeeded` / `VmnetNetworkRecord.reservedMACs`; deletion flow in `VMLibraryViewModel` never touches the store.

## Suggested fix

Replace the positional slot list with a persisted MAC → address map (addresses still assigned lowest-free-first, reassigned as a set when the subnet changes). That keeps address stability without an unreclaimable index, and lets VM deletion free its entry.

nicholas-lonsinger:
Closes #830.

## Summary

`VmnetNetworkService`'s per-network reservation table was append-only — `reserveAddressIfNeeded` was its only mutator. Nothing freed a slot when a VM was deleted, when its MAC was edited or regenerated (#833's Generate flow makes that unbounded rather than one-per-VM), or when its mode switched away from an app-managed network. Slot position *is* the address (`host 2 + index`), so each dead slot permanently costs the network an address its DHCP can no longer hand out dynamically — 253 of them on the /24 vmnet reserves.

Slots are now released **in place**: `VmnetNetworkRecord.reservedMACs` holds `[String?]`, where a `nil` element is a free slot the next MAC needing one refills. That reclaims the address without moving any surviving VM's, and needs no compaction. Trailing free slots are trimmed so an emptied table reads as empty.

Three release points:

| Trigger | Where |
|---|---|
| VM leaves the library (delete, external deletion, phantom cleanup) | `VMLibraryViewModel.evict` — the single removal path |
| MAC edited/regenerated, mode switched, networking turned off | `updateConfiguration`'s retirement block, beside the existing forwarding-rule withdrawal |
| Slots orphaned while the app was not running, and those already accumulated | `pruneAddressReservations` at load |

A release is guarded by a library-wide survivor check: duplicate MACs are legal today (#835), so only the last claimant frees the slot. The retirement release runs **before** `syncAddressReservation(for: new)`, mirroring the existing withdraw-then-declare order for rules — the freed slot is then the lowest one available, so an edited MAC normally keeps the VM's address.

## Deviations from the issue's suggested fix

The issue proposed replacing the positional list with a persisted MAC → address map. A sparse list was chosen instead:

- **A map breaks the persisted shape.** Under AGENTS.md's no-migration rule the old key would simply be ignored, so every existing install would lose `reservedMACs` and re-reserve in library order — reshuffling every guest's IP address once, which is exactly the stability the issue wanted to protect. `Optional<String>` decodes from a plain JSON string, so an existing `networks.json` reads unchanged and nothing migrates.
- **A map needs a reassignment path that does not exist.** `materialize` persists system-adjusted addressing, which would strand every stored address outside the new subnet.
- A `nil` hole reclaims the slot just as completely and leaves the positional derivation, capacity handling, and slot-ordered forwarding-rule dedupe exactly as they are.

Two additions beyond the issue: the survivor check (needed while duplicate MACs are legal) and the load-time prune (without it, slots already accumulated — the issue's headline complaint — are never reclaimed).

## Scope

Kept to #830. #834 (a reservation-pending network never rematerializing) and #835 (duplicate MACs) are untouched — no pending flag, no validation, no UI. Two notes for whoever takes them:

- **#834** — a release is now a second reason a materialized network's reservations differ from the record, so its pending comparison must be written against the whole reservation set, not just additions. Its deferred question ("can a retired MAC's slot be reused without moving any other VM's address?") is answered *yes* here: a freed slot keeps its index.
- **#835** — the survivor check is correct while duplicates are legal and stays correct once they are forbidden.

The issue carries `Type: Refactor`, but this changes behavior (slots are now freed and reused), so the commit uses the `fix` prefix.

## Test plan

- [x] `make test` — full suite green, including 17 new cases
- [x] `make lint`

New coverage: slot reuse without moving other addresses, case-insensitive release, a no-op release writing nothing, trailing-slot trimming, a free slot surviving a relaunch, retain semantics (including cross-network isolation), a reused slot installing for its new owner at the next materialization, and the view model's delete / MAC-edit ordering / mode-switch / networking-off / duplicate-MAC-survivor / load-prune / failed-bundle / unentitled paths.

### Maintenance Notes

- ✅ **docs/ARCHITECTURE.md** — skipped, no component boundary changed. Its description ("each VM holds a slot keyed on its persisted MAC … kept in step with configurations through `VMLibraryViewModel`'s persistence funnel") stays true; `VmnetNetworkProviding` gained two methods on an existing seam.
- ✅ **Tests** — 17 new cases across `VmnetNetworkServiceTests` and `VMLibraryViewModelTests`; `MockVmnetNetworkProvider` extended.
- ✅ **AGENTS.md** — skipped, no rule stated there changed.
- ✅ **RATIONALE:** — none added, so no `## Notes` disclosure beyond the commit's.

### DIFF
--- a/Kernova/Services/VmnetNetworkService.swift
+++ b/Kernova/Services/VmnetNetworkService.swift
@@ -9,7 +9,7 @@ import vmnet
 /// (docs/NETWORKING.md).
 // `CodingKeyRepresentable` so a `[VmnetNetworkKind: …]` store encodes as a
 // JSON object keyed by kind, not Codable's array-of-pairs fallback.
-enum VmnetNetworkKind: String, Codable, CodingKeyRepresentable, Sendable {
+enum VmnetNetworkKind: String, Codable, CodingKeyRepresentable, CaseIterable, Sendable {
     /// The Host Only network: guests on it reach the host and each other,
     /// never the LAN or the internet.
     case hostOnly
@@ -71,14 +71,18 @@ enum IPv4Value {
 }
 
 /// What one app-managed network persists across launches: its addressing once
-/// first reserved, and the ordered MACs holding DHCP reservation slots —
-/// slot order is the address assignment, so the list only ever appends.
+/// first reserved, and the ordered MACs holding DHCP reservation slots.
+///
+/// Slot position *is* the address (``VmnetNetworkAddressing/reservedAddress(slot:)``),
+/// so a slot is released by emptying it in place — a `nil` element is a free
+/// slot the next MAC needing one refills, leaving every surviving VM's address
+/// where it was. Trailing free slots are trimmed.
 struct VmnetNetworkRecord: Codable, Sendable, Equatable {
     /// `nil` until the network first materializes.
     var addressing: VmnetNetworkAddressing?
-    var reservedMACs: [String]
+    var reservedMACs: [String?]
 
-    init(addressing: VmnetNetworkAddressing? = nil, reservedMACs: [String] = []) {
+    init(addressing: VmnetNetworkAddressing? = nil, reservedMACs: [String?] = []) {
         self.addressing = addressing
         self.reservedMACs = reservedMACs
     }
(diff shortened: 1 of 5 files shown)


nicholas-lonsinger:
## Review pass

Three reviewers: `/code-review high`, plus the Codex review and adversarial-review lanes.

**Fixed on the branch** — `reconcileWithDisk` evicts a resting VM whose bundle is on disk but whose configuration failed to parse, and the new release meant eviction also freed its reservation slot. That is exactly the case `pruneAddressReservations(scanWasComplete:)` protects at load, so the two paths disagreed. `evict` now takes `bundleIsGone`, and the reconcile passes `false` for unreadable bundles. Raised by both `/code-review` and Codex; covers two new tests.

**Deferred to #837** — a retained MAC whose slot index sits past the subnet's capacity is never relocated into a freed hole, so the VM actually starved of an address cannot use the reclaimed capacity. Raised by both Codex lanes (P1 / high). Kept out of this PR because the state needs a table of 254+ accumulated slots, which #830 itself estimated at years of churn; the issue records the gap and leaves "not worth fixing" open as an outcome.

**Dismissed**, both under [docs/REVIEW.md](docs/REVIEW.md)'s "pre-existing behavior surfaced by an unrelated diff":

- `evict`'s port-forwarding withdrawal has no survivor check while the new reservation release does, so deleting one of two VMs sharing a MAC withdraws the survivor's rules. Verified against the merge base — the withdrawal is unchanged by this diff, and duplicate MACs need a hand-edited config. Belongs with #835.
- The retirement release runs before `saveConfiguration`, so a failed write leaves the reservation table ahead of disk. The adjacent `declarePortForwardingRules([], for: old)` has done the same on this path since #832; not introduced or worsened here.

Issue #837, opened 2026-08-14
Relocate a reservation slot whose index sits past the subnet's capacity


nicholas-lonsinger:
## Found during

PR #836 review (`/code-review high` plus both Codex lanes) of the slot-reclaim machinery. Flagged independently by two reviewers.

## Description

`retainAddressReservations` and `releaseAddressReservation` free a slot by emptying it in place, and `reserveAddressIfNeeded` refills the lowest free hole. Neither moves a MAC that is *retained*. That is deliberate — a surviving VM's address must not shift — but it leaves one state unrepaired.

A reservation table carried over from before the reclaim existed can hold a still-live MAC at an index past the subnet's last usable address. `VmnetNetworkAddressing.reservedAddress(slot:)` answers `nil` there, so `resolveSlots` classifies it `noAddressLeft`: the VM gets no DHCP reservation, its IP Address row shows the placeholder, and its port-forwarding rules cannot install (they are dropped as "a VM holding no address reservation"). Reclaiming earlier slots does not help it, because `reserveAddressIfNeeded` returns early on a MAC already present in the table.

So the reclaim frees capacity that the one VM actually starved of an address cannot use. Relocating such a MAC is safe in a way relocation generally is not: it currently has no address, so nothing moves out from under the guest.

## Location

`VmnetNetworkService.retainAddressReservations`, `VmnetNetworkService.reserveAddressIfNeeded`, `VmnetNetworkService.resolveSlots` — and `VmnetNetworkAddressing.reservedAddress(slot:)`, which defines the capacity.

## Considerations

Reachability is the open question, and it is the reason this was not folded into #836. On the /24 vmnet reserves the table has 253 usable slots, and #830 estimated that filling it takes years of clone/delete churn — so no install is expected to hold such a store. The finding is recorded because it is a verified completeness gap in the reclaim itself, and because whoever picks it up should decide whether the state is worth writing code for at all; closing it as not worth fixing is a legitimate outcome.

A fix also needs to say when the relocation runs (launch-time pruning is the natural point) and to cover the smaller-subnet variant, where addressing changes under an existing table rather than the table outgrowing it.

nicholas-lonsinger:
Closing as not worth fixing.

The gap is real and still present: `reserveAddressIfNeeded` returns early on a MAC already in the table without asking whether its slot resolves to an address, and neither release path moves a retained MAC, so `resolveSlots` keeps classifying a beyond-capacity slot `.noAddressLeft`.

[...]

PR #838, opened 2026-08-14
fix: Recreate a network when its DHCP reservations change

### LINKED ISSUE #834: A changed or new MAC's DHCP reservation never activates on an already-materialized network

nicholas-lonsinger:
## Summary

A MAC added, changed, or freed while its app-managed network was already materialized never reached guests. Reservations are fixed at network creation, so a slot assigned afterwards is pending until the next materialization — but nothing marked a *reservation* change as needing a recreate: the idle rebuild consulted only `portForwardingRulesArePending(for:)`, and it was hardcoded to Shared.

Viewing any Shared or Host Only VM's Network section materializes that network to fill the IP Address row, so the state is easy to reach: after that, editing a MAC (or creating, cloning, or importing a VM) left the row showing `—` for the rest of the app session and handed the guest a dynamic lease instead of its deterministic reservation. Host Only had no idle-recreate path at all.

## Approach

One predicate covers both declaration sets. `portForwardingRulesArePending(for:)` becomes `networkConfigurationIsPending(for:)`, which compares the materialized network's whole reservation set against what would install right now, then its forwarding rules. The sole consumer is the idle recreate, which does not care *which* set differs — a second sibling predicate would make the call site assemble the same disjunction.

Two properties the comparison has to hold:

- **Both sides are the *installable* sets.** A slot past the subnet's capacity, a slot holding a MAC that no longer parses, and a rule whose VM holds no reservation can never install; measuring against them would leave the flag stuck true and drive an endless recreate. A new `installableReservationsLocked(for:)` is the shared slot resolution that the rules derivation now reads too.
- **The whole set, not just its additions.** Since #836 a release is a second reason the materialized network differs from the record, so a departed VM's freed slot also pends — the live network keeps honoring that reservation until the recreate.

On the view model side, `rebuildSharedNetworkIfIdle()` becomes `rebuildNetworksIfIdle()` over every `VmnetNetworkKind`. The rebuild also moves off `declarePortForwardingRules` (now a pure declaration, with no hidden side effect) and onto the batch boundaries — `wirePersistence`, `updateConfiguration`, `evict`, `pruneAddressReservations`, and session teardown — so one pass of the funnel produces at most one recreate per kind. `evict` gained one on both of its paths; it previously returned before any rebuild when the bundle was still on disk.

A recreate only fires when no VM could be attached, and it keeps the network's pinned addressing, so no guest's address moves.

## Test plan

- [x] `make test` — 2809 tests, 0 failures (new test names verified present in the xcresult, not just the `TEST SUCCEEDED` banner)
- [x] `make lint`
- [x] Service tests: a slot taken, released, or refilled while materialized pends and stops after the recreate; a refilled slot takes the departed VM's address while every survivor keeps its own; a beyond-capacity slot never pends; nothing pends before the first materialization
- [x] View model tests: a changed MAC recreates Shared and Host Only, a mode switch recreates both exactly once, delete and library load recreate, and a live VM defers the recreate until its session is torn down

## Notes

Adds no `RATIONALE:` comments.

`MockVmnetNetworkProvider` swapped its single `isMaterialized` flag for per-kind `materializedKinds` state — one rebuild pass now queries every kind, and a shared flag would let the first invalidation mask the rest.

### Deviations from the issue

- The issue proposed exposing reservation-pending state *alongside* the forwarding-rules predicate; this replaces it with one predicate, per the reasoning above.
- The issue's closing paragraph ("`reservedMACs` is append-only and persisted, so a retired MAC keeps its slot forever… decide separately whether a retired MAC's slot can be reused") is stale — #836 already made the list sparse with in-place refill, so no slot-reuse decision remained open.

### Scope

- **#835** (duplicate-MAC collision) is unaffected: the comparison is MAC-keyed and both sides dedupe identically, so a duplicate MAC compares equal rather than pending. Whatever resolves the collision will trigger a recreate through this machinery.
- **#837** (relocating a beyond-capacity MAC into a freed hole) is deliberately *not* absorbed — beyond-capacity slots are excluded from the comparison precisely so they cannot drive a recreate loop, and nothing here moves a slot.

Closes #834

### DIFF
--- a/Kernova/Services/VmnetNetworkService.swift
+++ b/Kernova/Services/VmnetNetworkService.swift
@@ -339,10 +339,10 @@ protocol VmnetNetworkProviding: Sendable {
     /// network's next materialization.
     func setPortForwardingRules(_ rules: [PortForwardingRule], for mac: String, kind: VmnetNetworkKind)
     /// Whether the materialized network of `kind` carries a different set of
-    /// forwarding rules than the one that would install right now — so
-    /// recreating it would change what is forwarded. `false` while no network
-    /// of `kind` is materialized.
-    func portForwardingRulesArePending(for kind: VmnetNetworkKind) -> Bool
+    /// DHCP reservations or forwarding rules than the ones that would install
+    /// right now — so recreating it would change what guests get. `false`
+    /// while no network of `kind` is materialized.
+    func networkConfigurationIsPending(for kind: VmnetNetworkKind) -> Bool
     /// The kind whose materialized network `network` is, `nil` for a network
     /// this service does not hold.
     func kind(ofNetwork network: vmnet_network_ref) -> VmnetNetworkKind?
@@ -366,7 +366,7 @@ protocol VmnetNetworkProviding: Sendable {
 /// and rules can only be added to a configuration), so either one declared
 /// while its network is materialized takes effect at the next materialization
 /// — the next app launch, or a recreate driven by recovery or by
-/// ``portForwardingRulesArePending(for:)``.
+/// ``networkConfigurationIsPending(for:)``.
 ///
 /// Lock-guarded `Sendable` rather than `@MainActor`: it never touches
 /// `VZVirtualMachine`, and `ConfigurationBuilder` consumes it during off-main
@@ -625,7 +625,7 @@ final class VmnetNetworkService: @unchecked Sendable {
     /// Resolves every reservation slot in order, without logging — the shared
     /// source of truth for what a network of `addressing` would install, read
     /// both while materializing and while answering
-    /// ``portForwardingRulesArePending(for:)``.
+    /// ``networkConfigurationIsPending(for:)``.
     private static func resolveSlots(
         macs: [String?], addressing: VmnetNetworkAddressing
     ) -> [ResolvedSlot] {
@@ -751,6 +751,20 @@ final class VmnetNetworkService: @unchecked Sendable {
         return desiredForwardingRules[kind] ?? [:]
     }
 
+    /// The MAC → address reservations a network of `kind` created right now
+    /// would install, in slot order — slots that cannot install (past the
+    /// subnet's capacity, or holding a MAC that no longer parses) are not in
+    /// it, so it is directly comparable with `installedAddresses`.
+    ///
+    /// The caller holds `stateLock`.
+    private func installableReservationsLocked(
+        for kind: VmnetNetworkKind
+    ) -> [(mac: String, address: String)] {
+        guard let record = records[kind], let addressing = record.addressing else { return [] }
+        return Self.installablePairs(
+            Self.resolveSlots(macs: record.reservedMACs, addressing: addressing))
+    }
+
     /// The rules a network of `kind` created right now would carry, keyed by
     /// MAC — declarations that cannot install (no reservation slot, a host port
     /// another VM already claims) are not in it, so it is directly comparable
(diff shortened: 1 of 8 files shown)


nicholas-lonsinger:
## Review round

Three reviews ran over the branch: the built-in `/code-review high`, plus a Codex review and a Codex adversarial review. Findings and triage:

**Fixed now**

1. **The destination network was recreated too late on a live switch** (Codex review). Moving the rebuild off `declarePortForwardingRules` and onto the batch boundaries put it *after* `applyLivePolicy`. A live session reads as attached to the network it is *on*, so the network a VM switches *to* is idle only until the coordinator attaches it — synchronously. A running VM switching into an already-materialized Shared network therefore protected it before its own reservation and rules installed. `updateConfiguration` now rebuilds before `applyLivePolicy` as well as after, restoring the pre-apply pass the old code got for free.

2. **The teardown rebuild was dropped whenever the session ended mid-transition** (built-in review). `tearDownSession` fires `onSessionTornDown` *before* the caller settles the status, and `mayHoldAttachment` falls back to `status.isTransitioning` with no live `VZVirtualMachine` — so the VM that had just released the network read as still holding it. This hit save-suspend (`.saving`), which is how the app normally leaves a running VM including on quit, plus the cancelled-guest-setup (`.installing`) and terminal start-failure (`.starting`) teardowns. The scan now skips the instance whose session just ended.

   `MockVirtualizationService.save` did not set `.saving` before tearing down as the real service does, which is why the existing `saveSuspendReleasesTheSharedNetwork` could not observe this. With the mock corrected, both that test and the new `.saving` test fail if the guard is reverted — verified.

3. The recreate log line no longer enumerates causes; a Host Only recreate can never be rule-driven, since rules are only ever declared on Shared.

**Deferred — #839**

The adversarial review found that a reservation declared *while* its network is being created is missed: `network(for:)`'s pre-publish staleness check covers forwarding rules only. It does read as pending once the network publishes, so it converges on the next trigger rather than persisting to relaunch — a strictly narrower window than what this PR fixes, and a distinct defect. Filed as #839 rather than widening this PR; it also interacts with the system-adjusted-addressing path, which needs its own decision.

**Dismissed**

Nothing else. The reviews independently cleared the areas most at risk here: the pending predicate is symmetric on both sides (so no stuck-true recreate loop), `installableReservationsLocked` is only reached from `stateLock`-holding call sites (no re-entrant `NSLock`), every `declarePortForwardingRules` site is still covered by a rebuild in the same transaction, and `wirePersistence`'s rebuild running before the instance joins `instances` is safe at every construction site.

`make test`: 2810 tests, 0 failures. `make lint` clean.

Issue #839, opened 2026-08-14
A DHCP reservation declared while its network is being created is never installed


nicholas-lonsinger:
## Found during

Adversarial review of PR #838 (recreating a network when its DHCP reservations change).

## Description

`VmnetNetworkService.network(for:)` re-checks one declaration set before publishing a freshly created network: `Self.rulesByMAC(materialized.submittedForwardingRules) != installableForwardingRulesLocked(for:)`. A rule declared while the vmnet create round-trip is in flight is caught there, and the network is released and recreated (up to `materializeAttemptLimit`).

Reservations have no such check. A slot taken while a network is materializing is not in the network that comes back, and the publish path installs `installedAddresses[kind]` from what the create actually carried.

`networkConfigurationIsPending(for:)` answers `false` while no handle is published, so the rebuild pass that runs when the slot is taken (VM created, cloned, imported, or its MAC edited) finds nothing to do. Once the network publishes, the slot does read as pending — but nothing re-checks at that moment, so the recreate waits for an unrelated trigger: another configuration change, a VM deletion, a session teardown, or an app relaunch.

Until then the VM's IP Address row shows `—`, its guest takes a dynamic lease rather than its deterministic reservation, and any forwarding rules riding that reservation are dropped as unreserved.

The window is the vmnet create round-trip on a background thread, with the main actor free to accept the declaration.

## Location

`VmnetNetworkService.network(for:)`, `VmnetNetworkService.networkConfigurationIsPending(for:)`, `VmnetNetworkService.MaterializedNetwork.submittedForwardingRules`

## Suggested fix

Carry a submitted-reservations counterpart alongside `submittedForwardingRules` and fold it into the same pre-publish staleness check, so a reservation declared mid-create releases and recreates the network exactly as a rule does.

One interaction to settle when this is picked up: the system-adjusted-addressing path (`materialize`) deliberately publishes with `reservations: []` after rewriting the record's addressing. A reservation comparison would read that as stale on every attempt, so that path needs to be excluded or reconciled — `adjustedPinnedAddressingLeavesReservationsPending` covers the current behavior.

PR #840, opened 2026-08-14
fix: Refuse a MAC address another VM already holds

### LINKED ISSUE #835: Two VMs can hold the same MAC address, colliding on their reservation and forwarding keys

nicholas-lonsinger:
Closes #835

## Summary

Nothing checked a MAC address against the rest of the library, so two VMs could hold the same one and collide on both things keyed on it: the DHCP reservation slot (a single slot, both guests promised one address) and `desiredForwardingRules[kind][mac]` (last writer wins, and either VM's withdrawal disarmed the other's rules until something re-declared them).

The app now refuses the second holder. Enforcement sits at `updateConfiguration(of:mutate:)` — the single configuration funnel — so the MAC field, Generate, and any future writer are covered rather than one surface at a time. A refusal logs at `.notice`, surfaces an alert naming the VM that holds the address, and leaves the mutation entirely unapplied (including any other field the same closure set). The MAC field then shows the address the VM actually ended up with, so a refused address snaps back exactly like a malformed one, with the alert supplying the reason the snap-back alone never could.

Carrying an address across from a VM being replaced stays reachable, and the alert says how: change or delete the VM holding it first.

## The route taken

- **Refuse rather than re-key.** The alternative was re-keying the reservation table and `desiredForwardingRules` on VM identity so duplicates stop colliding. Rejected: two stations sharing an L2 identity on one segment is broken regardless of what the app does, so re-keying would make the app carefully support a configuration that cannot work — and it is a far wider change to a subsystem #836 and #838 both landed in this week. Refusing is what makes the MAC a legitimate key.
- **The surface is `surfaceError`, not an inline caption.** This matches `refuseIfDuplicateMachineIDConflict`, the repo's existing duplicate-identity refusal, and puts the reason at the same choke point as the enforcement — so every writer inherits it instead of each re-implementing one. No new view state or refresh lifecycle.
- **"Held" means any VM carrying the address**, including one with networking off. The address persists across mode changes, so a narrower test would let a VM with networking off keep a colliding address and re-form the collision the moment networking came back on.
- **`evict` and the MAC-retirement block no longer clear the shared key.** Both pushed `[]` through `desiredForwardingRules[.shared][mac]`; with a duplicate pair that silently disarmed the other VM's forwarding. `withdrawPortForwardingRules(for:)` re-declares a same-MAC survivor's own rules instead. This is the finding #836 dismissed into this issue's territory, and it is the pattern already used for reservations (`releaseAddressReservationIfUnused`) — leaving one of the two MAC-keyed teardowns unguarded would have been divergence.

## Duplicates already on disk

No load-time dedupe, no regeneration, no rejection at load — deliberately, and not only because of the no-migration rule. Rewriting a guest's MAC behind the user's back is a silent edit to something the guest can see and pin in its own network configuration, and which under Bridged the LAN's DHCP server may hold a reservation for; NETWORKING.md's north star points away from it. A pre-existing duplicate keeps behaving exactly as it does today, the survivor guard removes its only destructive edge, and the user repairs it by editing either VM onto a free address — which this change explicitly permits, since only a *change* of address is checked.

Import is untouched for the same reason, and is the remaining way a duplicate can enter the library (alongside a hand-edited `config.json`); the UUID short-circuit in `reserveAndImport` already catches re-importing the same VM.

## Queued issues

No material effect on #837 (relocating a beyond-capacity MAC into a freed hole) — that touches slot assignment, which this does not; it only reduces how many MACs can share a slot. None on #839 either, which compares forwarding rules only.

## Changes

- `VMLibraryViewModel`: `vmHoldingMACAddress(_:otherThan:)`, `refuseIfDuplicateMACAddress(_:on:)`, `withdrawPortForwardingRules(for:)`; the refusal wired into `updateConfiguration`, and both MAC-keyed teardowns routed through the survivor-guarded withdrawal.
- `VMSettingsViewController.applyMACAddressFieldEdit()`: the field follows the model after the write.
- `docs/NETWORKING.md`: new principle 9, "A MAC address belongs to one virtual machine" (the former 9 becomes 10).
- Tests: a `MAC Address Uniqueness` section (refusal, case-insensitivity, networking-off holder, whole-mutation drop, pre-existing duplicate still editable, both escape hatches), two survivor cases in `Port Forwarding Sync`, and a field-revert case in `VMSettingsViewControllerTests`.

## Test plan

- [x] `make test` — 2820 tests, 0 failures, totals read from the xcresult
- [x] `make lint`

### Maintenance Notes

- ✅ ARCHITECTURE.md — skipped: no component boundary, data flow, or actor isolation changed.
- ✅ Tests — 10 new cases covering the refusal, its escape hatches, and the survivor-guarded withdrawal.
- ✅ AGENTS.md — skipped: no rule stated there changed. The new rule constrains future networking decisions, so it lives in docs/NETWORKING.md.

### DIFF
--- a/Kernova/ViewModels/VMLibraryViewModel.swift
+++ b/Kernova/ViewModels/VMLibraryViewModel.swift
@@ -1525,8 +1525,8 @@ final class VMLibraryViewModel {
     }
 
     /// Frees `target`'s reservation slot unless a VM still in the library wants
-    /// the same one — duplicate MACs are possible, so the last claimant is what
-    /// releases the slot.
+    /// the same one — a bundle can arrive carrying an address another VM already
+    /// holds, so the last claimant is what releases the slot.
     private func releaseAddressReservationIfUnused(_ target: (kind: VmnetNetworkKind, mac: String)) {
         guard isVMNetworkingEntitled else { return }
         let stillWanted = instances.contains {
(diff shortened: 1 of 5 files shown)


nicholas-lonsinger:
Adversarial review raised that the refusal covers only the configuration funnel, so a bundle admitted by import, load, or reconcile can still arrive carrying an address another VM holds — and nothing refuses starting such a pair. Verified against `reserveAndImport`, whose only duplicate check is by VM UUID. Out of scope for #835 (the destructive edge is already removed here), filed as #841.

nicholas-lonsinger:
Review pass done — built-in `/code-review high` plus two Codex reviews.

**Fixed (second commit):**
- Generate committed an open field edit before writing, so typing a held address then clicking Generate succeeded with a fresh random address *and* raised an alert about an address no longer in play. It now discards the edit (`abortEditing()`), which is what "Generate overrides the edit" already meant.
- NETWORKING.md said an address could be freed by "clearing" it; emptying the field only snaps back, so the principle now names the two routes that exist and that the alert already gives.

**Deferred:** #841 — a bundle admitted by import, load, or reconcile can still carry a held address, and nothing refuses starting such a pair.

**Dismissed:** `evict(bundleIsGone: false)` keeps the VM's reservation slot but still withdraws its forwarding rules. Verified against the merge base — pre-existing, unchanged by this diff, and self-healing on the next successful reconcile before rules bind at network materialization.

PR #844, opened 2026-08-14
fix: Disclose a MAC address two VMs hold, and refuse the second start

### LINKED ISSUE #841: A bundle admitted by import or load can carry a MAC address another VM already holds
## Found during

Adversarial review of PR #840 (issue #835), which refuses a duplicate MAC address at `updateConfiguration(of:mutate:)`.

## Description

#840 makes the configuration funnel refuse a MAC address another VM in the library holds, which covers every writer inside the app: the MAC field, Generate, and anything future that edits an address. It deliberately does not touch the paths that admit a configuration the app did not author.

`reserveAndImport` reads the source bundle's configuration and registers the instance with the address it carries; its only duplicate check is by VM UUID, which short-circuits re-importing a VM already in the library but says nothing about the address. The library load and `reconcileWithDisk` likewise admit whatever each bundle holds. So a bundle whose MAC matches an existing VM's — hand-copied, hand-edited, or duplicated outside the app — enters the library and the two VMs share the one thing the address keys: a single DHCP reservation slot, so both guests are promised one IPv4 address, and one `desiredForwardingRules` entry on the shared network.

#840 removed the destructive edge here — a withdrawal now re-declares a same-MAC survivor's rules rather than clearing the shared key — so such a pair behaves as it did before rather than silently disarming each other's forwarding. What remains is that the app admits the configuration without saying anything.

Also open, and part of the same question: nothing refuses *starting* two VMs that share an address. The duplicate machine ID case has `refuseIfDuplicateMachineIDConflict` plus the `blockDuplicateMachineIDBoot` preference; there is no counterpart for a MAC.

## Location

`reserveAndImport`, `loadVMs`, `reconcileWithDisk`, `start` — `VMLibraryViewModel`. The enforcement this would extend is `refuseIfDuplicateMACAddress(_:on:)`.

## Considerations

- Design when picked up; the shape is genuinely open. Refusing the import outright blocks a legitimate operation with no in-app remedy (the VM never lands, so it cannot be edited onto a free address). Admitting it with a disclosure, or refusing only the start, both keep the bundle reachable.
- Silently regenerating the address is the one option #840 ruled out: the guest can see and pin its MAC, and under Bridged the LAN's DHCP server may hold a reservation for it.
- Whatever is chosen should cover load and reconcile as well as import, or a duplicate simply moves to the path that stays quiet.

nicholas-lonsinger:
## Summary

`reserveAndImport`, `loadVMs` and `reconcileWithDisk` admit whatever MAC address a bundle arrives carrying — their only duplicate check is by VM UUID. Two VMs then share the single DHCP reservation slot and the single `desiredForwardingRules` entry the address keys, and the IP Address row shows one address as fact for both. #840's funnel refusal covers every address the app itself writes, and deliberately did not reach these paths.

This takes the shape **admit, disclose, refuse the second start**:

- **Admission stays permissive** on all three paths. Refusing at the door leaves no in-app remedy — the MAC field cannot repair a bundle that was never admitted — and at load/reconcile it would hide a VM the user can see on disk. Tests pin this so it is not "fixed" back.
- **Disclosure** is a warning banner in the VM's Network section naming the other holder(s). Being a function of library state, it covers import, load and reconcile identically, with no launch-time modal. The Generate button directly above it is the one-click remedy.
- **Enforcement** sits where the collision actually bites: `start` and a cold `resume` refuse when a live VM shares the address *on the same network mode*, mirroring `refuseIfDuplicateMachineIDConflict`. A hot resume is not guarded — the live VM already holds its own address.
- **Diagnostics**: `logDuplicateMACAddressHolders()` records the pair at `.warning` wherever library membership changes, so a duplicate that arrived from disk is traceable.

The address is never silently regenerated (#840's reasoning stands: the guest can pin it, and under Bridged the LAN's DHCP server may hold a reservation for it).

`docs/NETWORKING.md` principle 9 claimed "the library refuses a second holder", which was only ever true of addresses the app authors; it now states the admitted-and-disclosed end state.

## Deviations from the issue

- **The issue's `## Location` named `reserveAndImport`, `loadVMs`, `reconcileWithDisk` as where enforcement would go; none of them gained a refusal** — only the diagnostic log — for the remedy reason above.
- **No `blockDuplicateMACBoot` preference**, deviating from the machine-ID precedent's other half. That toggle exists because *its* remedy is dangerous (regenerating a machine ID can stop a macOS 12 guest booting). The MAC remedy is one click on either VM, and #840 already refuses a duplicate address with no override — an escape hatch only at boot would make one invariant negotiable in one place and absolute in the other.
- **Refusal is narrowed to the same `networkMode`**, where the machine-ID guard needs no narrowing: a Shared VM and a Host Only VM sharing an address are on different networks and different reservation slots. Two bridged VMs are refused whatever interface each names — Automatic resolves at start, so which link they land on is not knowable in advance.
- **No alert on import.** A multi-bundle drop would alert per bundle, and the identical state arriving via load or reconcile cannot be alerted without a launch-time modal — so the banner was needed regardless. The machine-ID case is likewise silent at import and walls the user at start.

## Test plan

- [x] `make test` — 2846 tests, 0 failures, verified from the xcresult (`totalTestCount` 2846 / `failedTests` 0); 16 new cases
- [x] `make lint`

## Notes

- No `RATIONALE:` comments added.
- No overlap with the concurrent #837 work: nothing here touches `VmnetNetworkService`, and no new call into `VmnetNetworkProviding` is added — the guard and the banner read `instances` only.

Closes #841

### DIFF
--- a/Kernova/ViewModels/VMLibraryViewModel.swift
+++ b/Kernova/ViewModels/VMLibraryViewModel.swift
@@ -540,6 +540,13 @@ final class VMLibraryViewModel {
     }
 
     func start(_ instance: VMInstance, bootIntoRecovery: Bool = false) async {
+        // Ahead of the setup dispatch: guest setup builds and runs a
+        // `VZVirtualMachine` carrying the configured machine identity and MAC
+        // address, so a conflicting one must be refused before it reaches the
+        // installer, not only on the auto-boot that follows.
+        if refuseIfDuplicateMachineIDConflict(instance) { return }
+        if refuseIfDuplicateMACAddressConflict(instance, joining: instance.configuration) { return }
+
         // Dispatch on the surviving setup context, not status, so `.error`
         // retries route through the same pipeline too.
         if instance.configuration.installContext != nil {
@@ -550,9 +557,6 @@ final class VMLibraryViewModel {
             downloadAndAutoBoot(instance)
             return
         }
-
-        if refuseIfDuplicateMachineIDConflict(instance) { return }
-        if refuseIfDuplicateMACAddressConflict(instance) { return }
 
         surfaceDisplay(for: instance)
         applyMatchWindowBootResolution(to: instance)
@@ -627,14 +631,20 @@ final class VMLibraryViewModel {
     /// Refuses an operation that would put a second guest on one MAC address on
     /// one network, logging the refusal and surfacing the alert.
     ///
+    /// `config` is the configuration the VM would run under — its own at start,
+    /// the prospective one for a live mode switch, which moves an unchanged
+    /// address onto a different network.
+    ///
     /// ``refuseIfDuplicateMACAddress(_:on:)`` keeps the library unique for every
     /// address the app writes; this covers the pair a bundle arrived carrying,
     /// which passed through no writer.
     ///
     /// - Returns: `true` when the caller must abort.
-    private func refuseIfDuplicateMACAddressConflict(_ instance: VMInstance) -> Bool {
-        guard let conflict = liveMACAddressConflict(for: instance),
-            let mac = instance.configuration.macAddress
+    private func refuseIfDuplicateMACAddressConflict(
+        _ instance: VMInstance, joining config: VMConfiguration
+    ) -> Bool {
+        guard let conflict = liveMACAddressConflict(for: config, excluding: instance),
+            let mac = config.macAddress
         else { return false }
         Self.logger.notice(
             "Refused to run '\(instance.name, privacy: .public)': shares the MAC address \(mac, privacy: .public) with active VM '\(conflict.name, privacy: .public)'"
@@ -647,8 +657,8 @@ final class VMLibraryViewModel {
         return true
     }
 
-    /// The first live VM sharing `instance`'s MAC address on the network it
-    /// would join, if any.
+    /// The first live VM sharing `config`'s MAC address on the network `config`
+    /// joins, if any.
     ///
     /// Live means VZ holds the attachment, as it does for the machine identity.
     /// The mode names the network, so two holders collide only where both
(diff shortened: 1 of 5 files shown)


nicholas-lonsinger:
## Review round

Three reviewers (built-in `/code-review high`, Codex review, Codex adversarial review) converged on one defect, plus two more that were verified and fixed here:

- **The live Mode picker bypassed the guard** (all three reviewers). Two holders permitted to run on different modes could be switched onto one network: `updateConfiguration` checked duplicates only when the *address* changed, and a mode-only mutation hot-swaps the attachment. The funnel now refuses a mode change that moves a running VM onto a network a live holder of its address is already on, and the picker re-selects the configured mode. A VM already in a collision stays editable, so it can leave one.
- **Guest setup ran before the guard** (both Codex lanes). A start carrying an install context returned through `installAndAutoBoot` / `downloadAndAutoBoot` first, and setup builds and runs a `VZVirtualMachine` with the configured address — only the post-install auto-boot was checked. Both identity guards now sit ahead of the dispatch; the machine-ID guard moved with it rather than leaving two sibling guards enforcing the same class of invariant at different points.
- **Three near-identical MAC-holder lookups** (built-in). Collapsed onto `vmsHoldingMACAddress(_:otherThan:)`, with the two callers deriving from it.
- **`logDuplicateMACAddressHolders()`'s doc overclaimed its call sites** (built-in). Narrowed to the three admission paths it actually runs on.

Deferred as out of scope: **#845** — two holders of one address share a single forwarding-rule declaration, last writer wins. Real and verified, but resolving it is a forwarding-model decision (per-VM declarations, or an absolute single-holder start refusal) rather than something to bolt onto this PR. It is the pre-existing state issue #841 catalogued; this PR discloses it rather than changing it.

Dismissed: none — every finding raised was either fixed or filed.

PR #855, opened 2026-08-14
feat: Drag files onto the VM display to send them to the guest's Downloads

### LINKED ISSUE #298: feat: Drag-and-drop files onto the VM display to transfer into the guest
## Summary
Allow the user to drag files from the host (Finder) and drop them onto the running VM display to transfer them into the guest's Downloads folder.

## Current state
- The VM display view (`VMDisplayBackingView`, wrapping `VZVirtualMachineView`) accepts no drops. Drag-and-drop exists elsewhere: the sidebar (VM reorder / `.kernova` bundle import), the storage-disk boot-order sheet (row reorder only), and the clipboard window (`ClipboardDropContainerView`, `ClipboardEditorTextView`).
- Host↔guest file transfer already works through the clipboard subsystem over vsock (macOS guests): promise-based, paste-driven — files land wherever the user pastes in the guest. Nothing today saves incoming files to a fixed guest location or reveals them in Finder.
- Clipboard transfers already have progress UI on both sides: host-side `ClipboardProgressCenter` / `ClipboardIndicatorView`, guest-agent menu-bar progress ring plus `ClipboardProgressMenuItemView`. Visibility is time-gated by `ClipboardProgressTracker` (0.3 s reveal delay, 2 s idle linger); there is no size threshold. No user-facing cancel exists — `abort()` fires only on supersession, error, or disconnect.
- VirtioFS shared directories are a shipped, settings-configured feature (`VMConfiguration.sharedDirectories`, `ConfigurationBuilder`), but they are an always-on mount, not a per-drop transfer channel.
- The guest agent is macOS-only; Linux guests use the text-only SPICE clipboard path.

## Proposed behavior
- Register the VM display view as a dragging destination for file URLs.
- On drop, stream the files to the guest agent (reusing the clipboard streaming machinery), which writes them into the guest's **Downloads** folder.
- The guest agent shows transfer progress under the same rules as clipboard paste transfers — same time-gated reveal/linger, same progress-ring + menu presentation.
- Transfer progress should probably get a **Cancel** button. That would be the first user-facing cancel in the transfer machinery; decide whether it also extends to clipboard paste transfers rather than existing on the drag-and-drop path alone.
- When a transfer completes, the agent opens a Finder window on the Downloads folder.
- Agent availability gates the gesture:
  - If the agent has **never connected** for this VM, the display exposes no drag-and-drop behavior at all.
  - If the agent **was connected but currently isn't**, the display participates in the drag but rejects it the standard AppKit way — return an empty `NSDragOperation` from `draggingEntered`/`draggingUpdated`, so the pointer shows no accept badge and a release springs back to the origin. No custom rejection UI.
- Filename collisions in Downloads resolve by silent uniquing — never a prompt, never an overwrite — using Finder's "Keep Both" convention: `name 2.ext`, `name 3.ext`, … The gesture is a drag-copy into a folder and the flow ends in a Finder window, so names match what Finder itself would produce.

nicholas-lonsinger:
Closes #298

## Summary

Dragging files from Finder onto a running macOS VM's display transfers them into the guest's Downloads folder, uniqued the way Finder itself would, and reveals them selected in a Finder window when the drop finishes.

## The route taken

**A dedicated vsock channel, the same streaming engine.** The drop rides `KernovaVsockPort.drop` (49155) rather than the clipboard channel, because the clipboard listener only exists while clipboard sharing is on, its admission requires `clipboard.stream.v1`, and everything it delivers lands on the guest pasteboard. A drop has to work with sharing off and must not disturb either clipboard. The *byte plane* is reused wholesale: `DropOffer` is metadata only, the guest pulls each item with an ordinary `ClipboardRequest`, and the bytes cross on the existing `Begin`/`Chunk`/`End`/`Ack`/`Abort` path — so there is one streaming implementation, not two.

**Three availability states, decided in one place.** `VMInstance.displayDropAvailability`:

| State | Signal | Behavior |
|---|---|---|
| `.none` | `lastSeenAgentVersion == nil` (persisted) | the display is not a drag destination at all |
| `.disconnected` | agent seen before, channel or capability missing now | registered, returns the empty `NSDragOperation` — no accept badge, release springs back |
| `.available` | live drop channel + `drop.files.v1` | `.copy` |

**Cancel, everywhere or nowhere.** The issue asked whether a Cancel button should extend beyond drops. It does: `ClipboardProgressTracker.openSession(onCancelRequested:)` puts it on the readout itself, so the host's status item and the guest agent's both get it, for drops and clipboard transfers alike. `ClipboardProgressCenter.cancelCurrent()` routes a host Cancel to the VM whose transfer is actually on screen rather than broadcasting. Completed files stay in Downloads; the in-flight partial is deleted by the existing abort path; the offer survives, so a paste can pull again.

**Guest-side landing.** Files stage to temp and are *renamed* into `~/Downloads` (same volume, so nothing payload-scaled happens), named by a new `FinderStyleUniquing` (`report 2.pdf`, not staging's invisible `report (2).pdf`). Jobs run FIFO on a dedicated serial queue, so a blocking pull never holds the agent's main thread.

## Rejected alternatives

- **Carrying drops on the clipboard channel** — would make the feature depend on a clipboard toggle it has nothing to do with, and would have to keep drop payloads off the pasteboard by convention rather than by construction.
- **Copying the consume loop per channel.** The design note called for copying `VsockClipboardService`'s frame-routing switch (including its abort-ordering fix) into the drop service. That is four copies of one rule; `ClipboardStreamRouting` now holds it once and all four endpoints use it, with the host/guest difference in sender-abort delivery an explicit parameter rather than a silently divergent copy.
- **A new `ClipboardTransferIssue.Kind` case for drop failures** — every other flow uses a factory over `.localFailure`, and a new case would have forced changes into an exhaustive `switch` that has nothing to do with drops.
- **A drop-only Cancel** — see above.

## Deviations from the design note

Verified against HEAD before implementing; these are where the note's factual claims did not hold:

1. **Guest enablement hook.** The note reused `VsockGuestControlAgent.onStateChange` to wake the drop client on `Hello`. That callback is change-gated on a state already set to `.connected` *before* any `Hello` arrives, so it never fires for the capabilities — the drop client would have stayed paused forever. Added an explicit `onHostCapabilitiesChanged` hook, fired on `Hello` and on the per-connection reset.
2. **`ClipboardProgressCenter` held no source reference.** The note assumed `progressBySource` already kept one; it keyed bare `ObjectIdentifier`s with no reference at all. Added a weak-source entry plus winner tracking.
3. **Guest-inbound clipboard cancel does not exist to wire.** The guest agent opens no inbound progress session (an inbound paste blocks its main thread, so no readout could repaint or be clicked). The same reasoning applies to the host's paste-time blocking pull, so neither gets a cancel closure; the sessions that *can* be cancelled are the two outbound ones plus both drop directions.
4. **One shared guest tracker required a teardown fix.** Sharing a tracker between the clipboard and drop agents made `clearAll()` on a clipboard teardown wipe a live drop's readout. The clipboard agent now closes only its own session.
5. **Peer-supplied text is not rendered.** `DropComplete` carries a machine-readable `code` alongside its message; the host maps the code to a sentence it composes itself, and the guest's message is logged only.
6. **No pbxproj source-list edits.** The targets use synchronized folder groups, so new files are picked up automatically.
7. `featureChannelAdmission` was refactored to take a `FeatureChannelRequirement` (existing two call sites migrated), and `startVsockServices()` needed no OS branch — `ConfigurationBuilder` only creates a socket device for macOS guests.

## Changes

- `KernovaKit/Proto/kernova.proto` — `DropOffer`/`DropComplete`/`DropRelease` at 40-42; regenerated bindings.
- `KernovaCapability.dropFilesV1`, `KernovaVsockPort.drop` (host + agent mirrors), three `ClipboardErrorCode` drop codes.
- New: `ClipboardStreamRouting`, `FinderStyleUniquing` (KernovaKit); `VsockDropService` (host); `VsockGuestDropAgent` (agent).
- Cancel plumbing across `ClipboardProgressTracker`, `ClipboardProgressSnapshot`, `ClipboardProgressMenuItemView`, `ClipboardProgressStatusItemPresenter`, `ClipboardProgressCenter`, both status-item controllers, and both clipboard endpoints.
- `VMDisplayBackingView` drag destination + wiring in `DetailContainerViewController` and `VMDisplayWindowController`.
- Agent `MARKETING_VERSION` 0.61.1 → 0.62.0 (both configurations) — the guest needs a reinstall for the new client.
- `docs/ARCHITECTURE.md` — drop service in the vsock stack; the agent now holds four connections.

## Test plan

- [x] `make test` — 2955 host + 170 agent + 408 package tests, all green
- [x] `make lint`
- [ ] Drag files and a folder onto a running macOS guest's display; confirm they land in Downloads, unique Finder-style against existing names, and are revealed selected
- [ ] Drag onto a VM whose guest agent is stopped; confirm no accept badge and a spring-back
- [ ] Cancel a large drop from the menu-bar readout; confirm completed files stay and the partial does not

## Notes

New coverage: `FinderStyleUniquingTests`, `ClipboardStreamRoutingTests`, `DropFrameTests`, `ClipboardProgressMenuItemViewTests`, `VsockDropServiceTests`, `VMInstanceDisplayDropTests`, `VsockGuestDropAgentTests`, plus additions to the progress-tracker, progress-center, display-backing-view, admission and control-agent suites.

### Maintenance Notes

- ✅ **ARCHITECTURE.md** — updated: a new service and a new guest agent are component boundaries; the vsock stack list and the agent's connection count now say so.
- ✅ **Tests** — every new public type and both new endpoints are covered, including the cancel paths and the three-state availability matrix.
- ⚠️ **AGENTS.md** — skipped: no rule stated there changed.
- ✅ **Follow-up filed** — the wire abort codes are still bare string literals matched across four files; typing them is out of scope here and is filed separately.

### DIFF
--- a/Kernova.xcodeproj/project.pbxproj
+++ b/Kernova.xcodeproj/project.pbxproj
@@ -829,7 +829,7 @@
 					"@loader_path/../Frameworks",
 				);
 				MACOSX_DEPLOYMENT_TARGET = 12.0;
-				MARKETING_VERSION = 0.61.1;
+				MARKETING_VERSION = 0.62.0;
 				PRODUCT_BUNDLE_IDENTIFIER = app.kernova.macosagent;
 				PRODUCT_MODULE_NAME = KernovaMacOSAgent;
 				PRODUCT_NAME = "Kernova Guest Agent";
@@ -867,7 +867,7 @@
 					"@loader_path/../Frameworks",
 				);
 				MACOSX_DEPLOYMENT_TARGET = 12.0;
-				MARKETING_VERSION = 0.61.1;
+				MARKETING_VERSION = 0.62.0;
 				OTHER_CODE_SIGN_FLAGS = "--timestamp";
 				PRODUCT_BUNDLE_IDENTIFIER = app.kernova.macosagent;
 				PRODUCT_MODULE_NAME = KernovaMacOSAgent;
--- a/Kernova/App/DetailContainerViewController.swift
+++ b/Kernova/App/DetailContainerViewController.swift
@@ -156,6 +156,16 @@ final class DetailContainerViewController: NSViewController {
             else { return }
             Task { await viewModel.resume(target) }
         }
+        backing.dropAvailability = { [weak viewModel] in
+            viewModel?.instances.first(where: { $0.id == instanceID })?.displayDropAvailability
+                ?? .none
+        }
+        backing.onDropFiles = { [weak viewModel] urls in
+            guard let target = viewModel?.instances.first(where: { $0.id == instanceID })
+            else { return false }
+            return target.sendDroppedFilesToGuest(urls)
+        }
+        backing.applyDropRegistration()
 
         backing.translatesAutoresizingMaskIntoConstraints = false
         view.addSubview(backing)
@@ -217,6 +227,12 @@ final class DetailContainerViewController: NSViewController {
                         _ = inst.virtualMachine
                         _ = inst.displayMode
                         _ = inst.configuration.displayAutoResizes
+                        // Everything `displayDropAvailability` reads, so the
+                        // display registers and unregisters as a drag
+                        // destination when the guest agent comes and goes.
+                        _ = inst.configuration.lastSeenAgentVersion
+                        _ = inst.vsockDropService?.isConnected
+                        _ = inst.vsockControlService?.isConnected
                     }
                 }
             },
@@ -242,6 +258,7 @@ final class DetailContainerViewController: NSViewController {
             guard let instance = viewModel.instances.first(where: { $0.id == id }) else { continue }
             backing.apply(
                 automaticallyReconfiguresDisplay: instance.configuration.displayAutoResizes)
+            backing.applyDropRegistration()
         }
 
         // Evict backing views for VMs no longer running inline
--- a/Kernova/App/HostAgentStatusItemController.swift
+++ b/Kernova/App/HostAgentStatusItemController.swift
@@ -47,7 +47,8 @@ final class HostAgentStatusItemController: NSObject, NSMenuDelegate {
     /// reaches the reattached menu rather than the popover's dismissal handler.
     private lazy var pasteProgressPresenter = ClipboardProgressStatusItemPresenter(
         statusItem: statusItem, menu: menu,
-        willAutoOpen: { [weak self] in self?.transientPopover.dismiss() })
+        willAutoOpen: { [weak self] in self?.transientPopover.dismiss() },
+        onCancel: { ClipboardProgressCenter.shared.cancelCurrent() })
 
     /// The status item's one transient-popover slot, shared by the soft-quit
     /// reminder and the clipboard notice.
--- a/Kernova/App/VMDisplayWindowController.swift
+++ b/Kernova/App/VMDisplayWindowController.swift
@@ -64,6 +64,13 @@ final class VMDisplayWindowController: NSWindowController, NSWindowDelegate {
 
         let backing = VMDisplayBackingView()
         backing.onResume = onResume
+        backing.dropAvailability = { [weak instance] in
+            instance?.displayDropAvailability ?? .none
+        }
+        backing.onDropFiles = { [weak instance] urls in
+            instance?.sendDroppedFilesToGuest(urls) ?? false
+        }
+        backing.applyDropRegistration()
         backing.update(
             virtualMachine: instance.virtualMachine,
             isPaused: instance.status == .paused,
@@ -190,6 +197,12 @@ final class VMDisplayWindowController: NSWindowController, NSWindowDelegate {
                 _ = self.instance.displayMode
                 _ = self.instance.configuration.clipboardSharingEnabled
                 _ = self.instance.configuration.displayAutoResizes
+                // Everything `displayDropAvailability` reads, so the display
+                // registers and unregisters as a drag destination when the guest
+                // agent comes and goes.
+                _ = self.instance.configuration.lastSeenAgentVersion
+                _ = self.instance.vsockDropService?.isConnected
+                _ = self.instance.vsockControlService?.isConnected
             },
             apply: { [weak self] in
                 guard let self else { return }
@@ -204,6 +217,7 @@ final class VMDisplayWindowController: NSWindowController, NSWindowDelegate {
                         transitionText: status.transitionLabel,
                         automaticallyReconfiguresDisplay: self.instance.configuration.displayAutoResizes
                     )
+                    self.backingView.applyDropRegistration()
                     self.updateToolbarItems()
                 }
             }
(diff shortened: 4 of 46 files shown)


nicholas-lonsinger:
## Review round

Three reviews ran on the first push (`/code-review high`, plus Codex's standard and adversarial passes). All three independently found the same defect, which is now fixed along with four others in a855cdf.

**Fixed**

- **The drop service never settled when its channel ended** (all three reviewers). `isConnected` stayed true on a dead channel, so `displayDropAvailability` kept returning `.available`: the display showed the accept badge and took a drop it could no longer send. The trigger is routine rather than rare — the guest's drop client pauses on every control reconnect, and `VsockGuestClient.pause()` closes the channel. The consume loop now settles the service on exit, sharing one idempotent path with `stop()`.
- **Cancel stopped one file, not the operation.** The peer decides what it pulls, so after `sender.cancel` it simply requested the next representation of a still-live offer and the readout came back; the host's own preview loop walked to its next rep for the same reason. Both sides now latch the cancelled generation and refuse further requests for it as `request.stale` — a code both ends already retire quietly, so the peer's paste comes back empty with nothing reported as a failure.
- **A folder still being sized when the channel went was silently discarded.** `startDrop` had already returned `true` and no job was registered yet, so `settle()` had nothing to report. That path now raises the interrupted-drop issue.
- **`dropFailed(code:)` threw away the specific code**, so the two drop-specific status-item lines were unreachable and a disk-full drop read as the generic "the files didn't reach the VM" in the dropdown while the notice body said otherwise.
- **The failure copy claimed nothing was saved.** The guest moves each file as it lands, so a batch failing partway leaves earlier files in Downloads; the wording now says the drop didn't finish.
- **`DropOffer`'s proto comment** claimed `is_inline` is always false. It isn't — an image file's rep carries its own flag — so the comment now says the field is ignored on this channel instead.
- **A finished drop job was cleared by generation without an identity check**, so a worker outliving a teardown could evict the next connection's job of the same number.

Each fix has a regression test; the suite and CI are green.

**Not changed, with reasoning**

- **A live-paused VM still accepts drops** (Codex adversarial, medium). Deliberate, and the same behavior the clipboard path has today: the offer is queued and the guest lands the files when it resumes, so the drop completes rather than being lost, and if the VM never resumes the teardown raises the interrupted-drop issue. The display is also showing its Paused overlay at the time. Refusing instead would remove a case that currently works.
- **"Defer landing until the whole batch succeeds"** (Codex adversarial). That means holding the batch in staging until the last byte, which is exactly the payload-scaled intermediate docs/CLIPBOARD.md §2 rules out. The wording fix addresses the honesty problem the finding is really about; reporting *which* files landed would need a new proto field and is not worth one.
- **Typing the wire abort codes.** Real, but pre-existing and repo-wide (~36 literal sites across four files); filed as #856 rather than expanded into this PR.
- **Whether the Cancel button inside a disabled `NSMenuItem` receives clicks** — not a code finding; it is on this PR's manual test plan, which is where it belongs.

